PR #3, opened 2019-09-11
FIX: Freezed time used in update_holiday_usernames_spec.rb should be UTC

lis2:
I was curious why this spec was failing on my local machine.

The reason for that was `from_date` of the post which in this spec was `2018-06-05 10:20:00 UTC`

Because I am in Australia timezone, `freeze_time Time.new(2018, 6, 5, 18, 40)` was giving me `2018-06-05 08:40:00 UTC` which was before `from_date`. 

I think that bug was introduced here: https://github.com/discourse/discourse-calendar/commit/b1fe4456d083be89188bcd57d3f4247a971637f5#diff-b1651fed94990b46f84a76daf7330e2dL17


SamSaffron:
Oh I see, this makes sense! we have lots of people very very close to UTC.

PR #5, opened 2019-11-21
FEATURE: Switch to using core user_options.timezone

martin-brennan:
See https://github.com/discourse/discourse/pull/8380, which must be merged before this.

* Remove group user timezone serializer, as this is now handled in core
* Delete connector and template for user-timezone, this is also now in the core user profile UI
* Use user_options.timezone instead of custom field


martin-brennan:
@eviltrout thanks Robin, after some thought I will actually change this to accept both flows. I set it this way in the first place due to discussion on dev, but I think we can have the best of both worlds. For the ruby side I can check `UserOption.column_names.include?('timezone')` and for the ember component I can just hide the whole component if `currentUser.user_option.timezone != undefined`

martin-brennan:
@eviltrout I've now changed it to respect both ways of doing things for a deprecation period. I show this warning now if `user_options.timezone` is present for `currentUser` in JS (not sure if there is a more subtle style), otherwise if a user of this plugin upgrades it BEFORE they upgrade core they should see no difference.

image

PR #6, opened 2019-12-10
(NO MERGE UNTIL EOF JAN) DEPRECATION: Completely remove user custom field timezone

martin-brennan:
The functionality of using a `UserCustomField` timezone for this plugin has been deprecated for the past 2 weeks and existing timezone custom fields have been migrated to the replacement, `UserOption.timezone`.


eviltrout:
I wonder if 2 weeks has been enough for all the people using it to update their installs. Do we have any idea of how many people are using this plugin?

martin-brennan:
@eviltrout I honestly have no idea how many people use this, or even if there is any way to find out. I'm happy to sit on this for longer. Anyone who has updated core discourse to a greater version than https://github.com/discourse/discourse/commit/afb553358162d4cbc2e6a1c01070a5420357d387 will be safe because the `UserOption` model and timezone migration is done there.

I'm happy to sit on this for a while longer to be safe?

gschlager:
Will it break for people using the stable branch of Discourse? I guess it would be okayish if some functionality starts to get missing, but it shouldn't completely break an install.

SamSaffron:
OK let's just wait till next stable release in Jan and nuke this then.

martin-brennan:
@gschlager hmm that is a good point, I'm not sure if the `UserOption.timezone` change is in stable yet. Will leave until end of Jan with a reminder as @SamSaffron suggests until merging

PR #8, opened 2019-12-16
FEATURE: Auto-trash direct replies for post with event that is trashed

martin-brennan:
* When the post with the event gets auto-trashed after the event has passed
* BUT do not trash posts that are direct replies IF they have their own events

### DIFF
--- a/jobs/scheduled/ensure_expired_event_destruction.rb
+++ b/jobs/scheduled/ensure_expired_event_destruction.rb
@@ -37,8 +37,6 @@ module Jobs
           next if (to_time + delay.hour) > Time.zone.now
 
           if post = pcf.post.topic.posts.find_by(post_number: post_number)
-            destroy_post(post)
-
             post.post_replies.each do |post_reply|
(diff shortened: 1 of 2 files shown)


jobs/scheduled/ensure_expired_event_destruction.rb
```diff
@@ -28,19 +28,31 @@ def execute(args)
...
           next if recurring
 
           to_time = to ? Time.parse(to) : Time.parse(from) + 24.hours
+          next if (to_time + delay.hour) > Time.zone.now
 
-          if (to_time + delay.hour) < Time.zone.now
-            if post = pcf.post.topic.posts.find_by(post_number: post_number)
-              PostDestroyer.new(Discourse.system_user, post).destroy
+          if post = pcf.post.topic.posts.find_by(post_number: post_number)
+            destroy_post(post)
```
jjaffeux: I'm just wondering, wouldn't be cleaner to try to destroy replies before destroying parent?

jjaffeux:
LGTM :+1: thanks

PR #11, opened 2019-12-20
FEATURE: Add timezone picker and options

martin-brennan:
Added these options for the markdown block, which utilise fullcalendar's inbuilt timezone functionality:

* `defaultTimezone` - Provide a default timezone to display the calendar dates in. If this is null we fall back to the current user timezone, and if that is null (or the current user is null) we fall back to moment.tz.guess(). This value is validated.
* `tzPicker` - If set to true, show a dropdown list allowing the user to pick the timezone the calendar displays in. The selected value is based on the rules described in tzdefault. If this is not turned on we show the text of the timezone in the header. The options for the picker come from moment.tz.names() and are rendered into the select at runtime, so we are not storing huge amounts of <option> HTML elements in the DB for no real reason.

I also hid the fullcalendar title because the title is now up in the header of the calendar.

With the picker enabled:

image

Without the picker enabled:

image

### DIFF
--- a/assets/javascripts/initializers/discourse-calendar.js.es6
+++ b/assets/javascripts/initializers/discourse-calendar.js.es6
@@ -313,6 +313,10 @@ function initializeDiscourseCalendar(api) {
     let defaultTimezone = null;
     if ($timezonePicker.length) {
       defaultTimezone = $timezonePicker.attr("data-default-timezone");
+      const isValidDefaultTimezone = !!moment.tz.zone(defaultTimezone);
+      if (!isValidDefaultTimezone) {
+        defaultTimezone = null;
+      }
     }
     const currentUserTimezone = currentUser ? currentUser.timezone : null;
     return defaultTimezone || currentUserTimezone || moment.tz.guess();
--- a/assets/javascripts/initializers/discourse-calendar.js.es6
+++ b/assets/javascripts/initializers/discourse-calendar.js.es6
@@ -46,8 +46,8 @@ function initializeDiscourseCalendar(api) {
   function render($calendar, post) {
     $calendar = $calendar.empty();
 
-    const timeZone = _getTimeZone($calendar, api.getCurrentUser());
-    const calendar = _buildCalendar($calendar, timeZone);
+    const timezone = _getTimeZone($calendar, api.getCurrentUser());
+    const calendar = _buildCalendar($calendar, timezone);
     const isStatic = $calendar.attr("data-calendar-type") === "static";
 
     if (isStatic) {
@@ -59,7 +59,7 @@ function initializeDiscourseCalendar(api) {
       _setDynamicCalendarOptions(calendar, $calendar);
     }
 
-    _setupTimezonePicker(calendar, timeZone);
+    _setupTimezonePicker(calendar, timezone);
   }
 
   function attachCalendar($elem, helper) {
@@ -75,7 +75,7 @@ function initializeDiscourseCalendar(api) {
   }
 
   function _buildCalendar($calendar, timeZone) {
-    let $calendarTitle = $(
+    let $calendarTitle = document.querySelector(
       ".discourse-calendar-header > .discourse-calendar-title"
     );
     const defaultView = escapeExpression(
@@ -83,7 +83,7 @@ function initializeDiscourseCalendar(api) {
     );
 
     return new window.FullCalendar.Calendar($calendar[0], {
-      timeZone: timeZone,
+      timeZone,
       timeZoneImpl: "moment-timezone",
       nextDayThreshold: "06:00:00",
       displayEventEnd: true,
@@ -108,9 +108,7 @@ function initializeDiscourseCalendar(api) {
         center: "title",
         right: "month,basicWeek,listNextYear"
       },
-      datesRender: info => {
-        $calendarTitle.text(info.view.title);
-      }
+      datesRender: info => ($calendarTitle.innerText = info.view.title)
     });
   }
 
@@ -315,25 +313,28 @@ function initializeDiscourseCalendar(api) {
       defaultTimezone = null;
     }
 
-    const currentUserTimezone = currentUser ? currentUser.timezone : null;
-    return defaultTimezone || currentUserTimezone || moment.tz.guess();
+    return (
+      defaultTimezone ||
+      (currentUser && currentUser.timezone) ||
+      moment.tz.guess()
+    );
   }
 
-  function _setupTimezonePicker(calendar, timeZone) {
+  function _setupTimezonePicker(calendar, timezone) {
     let $timezonePicker = $(".discourse-calendar-timezone-picker");
 
     if ($timezonePicker.length) {
-      $timezonePicker.on("change", function(e) {
-        calendar.setOption("timeZone", $(this).val());
+      $timezonePicker.on("change", function(event) {
+        calendar.setOption("timeZone", event.target.value);
       });
 
       moment.tz.names().forEach(timezone => {
         $timezonePicker.append(new Option(timezone, timezone));
       });
 
-      $timezonePicker.val(timeZone);
+      $timezonePicker.val(timezone);
     } else {
-      $(".discourse-calendar-timezone-wrap").text(`${timeZone}`);
+      $(".discourse-calendar-timezone-wrap").text(timezone);
     }
   }
 }
--- a/assets/javascripts/initializers/discourse-calendar.js.es6
+++ b/assets/javascripts/initializers/discourse-calendar.js.es6
@@ -46,7 +46,7 @@ function initializeDiscourseCalendar(api) {
   function render($calendar, post) {
     $calendar = $calendar.empty();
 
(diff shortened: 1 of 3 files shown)


assets/javascripts/initializers/discourse-calendar.js.es6
```diff
@@ -71,13 +74,16 @@ function initializeDiscourseCalendar(api) {
...
+    let $calendarTitle = $(
+      ".discourse-calendar-header > .discourse-calendar-title"
+    );
     const defaultView = escapeExpression(
       $calendar.attr("data-calendar-default-view") || "month"
     );
 
     return new window.FullCalendar.Calendar($calendar[0], {
-      timeZone: moment.tz.guess(),
+      timeZone: timeZone,
```
jjaffeux:
```suggestion
      timeZone,
```

assets/javascripts/lib/discourse-markdown/discourse-calendar.js.es6
```diff
@@ -1,30 +1,81 @@
...
-      token.attrs.push(["data-hidden-days", info.attrs.hiddenDays]);
+      mainCalendarDivToken.attrs.push([
+        "data-hidden-days",
+        info.attrs.hiddenDays
+      ]);
     }
+  },
+
+  after: function(state) {
+    state.push("div_close", "div", -1);
```
jjaffeux: you can probably give better names to div_open/div_close instead of comments

assets/javascripts/lib/discourse-markdown/discourse-calendar.js.es6
```diff
@@ -1,30 +1,81 @@
...
+
 export function setup(helper) {
   helper.whiteList([
     "div.calendar",
+    "div.discourse-calendar-header",
+    "div.discourse-calendar-wrap",
+    "select.discourse-calendar-timezone-picker",
+    "span.discourse-calendar-timezone-wrap",
+    "h2.discourse-calendar-title",
+    "option",
```
jjaffeux: might be worth to add a class so we don't have to whitelist so widely

assets/stylesheets/common/discourse-calendar.scss
```diff
@@ -106,3 +111,31 @@ a.holiday {
...
+  padding: 0.5em;
+  border-bottom: 1px solid $primary-low;
+  background: $primary-very-low;
+
+  .discourse-calendar-timezone-picker {
+    margin-bottom: 0;
+  }
+
+  .discourse-calendar-title {
+    margin-top: 10px;
```
jjaffeux: That's personal preference, but:

```suggestion
    margin-top: 1em;
```

assets/javascripts/initializers/discourse-calendar.js.es6
```diff
@@ -298,6 +307,35 @@ function initializeDiscourseCalendar(api) {
...
+    const currentUserTimezone = currentUser ? currentUser.timezone : null;
+    return defaultTimezone || currentUserTimezone || moment.tz.guess();
+  }
+
+  function _setupTimezonePicker(calendar, timeZone) {
+    let $timezonePicker = $(".discourse-calendar-timezone-picker");
+
+    if ($timezonePicker.length) {
+      $timezonePicker.on("change", function(e) {
+        calendar.setOption("timeZone", $(this).val());
```
jjaffeux: would `e.target.value` work instead of `$(this).val()` ?

assets/javascripts/initializers/discourse-calendar.js.es6
```diff
@@ -298,6 +307,35 @@ function initializeDiscourseCalendar(api) {
...
+      });
+
+      moment.tz.names().forEach(timezone => {
+        $timezonePicker.append(new Option(timezone, timezone));
+      });
+
+      $timezonePicker.val(timeZone);
+    } else {
+      $(".discourse-calendar-timezone-wrap").text(`${timeZone}`);
```
jjaffeux:
```suggestion
      $(".discourse-calendar-timezone-wrap").text(timeZone);
```

assets/javascripts/initializers/discourse-calendar.js.es6
```diff
@@ -298,6 +307,35 @@ function initializeDiscourseCalendar(api) {
...
+    const isValidDefaultTimezone = !!moment.tz.zone(defaultTimezone);
+    if (!isValidDefaultTimezone) {
+      defaultTimezone = null;
+    }
+
+    const currentUserTimezone = currentUser ? currentUser.timezone : null;
+    return defaultTimezone || currentUserTimezone || moment.tz.guess();
+  }
+
+  function _setupTimezonePicker(calendar, timeZone) {
```
jjaffeux: it's weird to me that we are writing `timeZone`  here when everywhere else is `timezone`

assets/javascripts/initializers/discourse-calendar.js.es6
```diff
@@ -71,13 +74,16 @@ function initializeDiscourseCalendar(api) {
     ).then(() => render($calendar, helper.getModel()));
   }
 
-  function _buildCalendar($calendar) {
+  function _buildCalendar($calendar, timeZone) {
+    let $calendarTitle = $(
```
jjaffeux: we don't need jquery for this, we only change `innerText` after

assets/javascripts/initializers/discourse-calendar.js.es6
```diff
@@ -101,6 +107,9 @@ function initializeDiscourseCalendar(api) {
         left: "prev,next today",
         center: "title",
         right: "month,basicWeek,listNextYear"
+      },
+      datesRender: info => {
```
jjaffeux: wonder if you can go without { } here

assets/stylesheets/common/discourse-calendar.scss
```diff
@@ -1,5 +1,6 @@
 .calendar {
   height: 690px; // Must be fixed to prevent height change on load
+  padding: 0 10px;
```
jjaffeux:
```suggestion
  padding: 0 0.5em;
```

assets/javascripts/initializers/discourse-calendar.js.es6
```diff
@@ -298,6 +307,35 @@ function initializeDiscourseCalendar(api) {
...
+
+  function _getTimeZone($calendar, currentUser) {
+    let defaultTimezone = $calendar.attr("data-calendar-default-timezone");
+    const isValidDefaultTimezone = !!moment.tz.zone(defaultTimezone);
+    if (!isValidDefaultTimezone) {
+      defaultTimezone = null;
+    }
+
+    const currentUserTimezone = currentUser ? currentUser.timezone : null;
+    return defaultTimezone || currentUserTimezone || moment.tz.guess();
```
jjaffeux: Maybe this instead of the two lines, especially given first line will always get called even if we have a defaultTimezonz?

```suggestion
    return defaultTimezone || (currentUser && currentUser.timezone) || moment.tz.guess();
```

assets/javascripts/lib/discourse-markdown/discourse-calendar.js.es6
```diff
@@ -1,30 +1,81 @@
...
+    // end div.discourse-calendar-header
+
+    // div.calendar
+    let mainCalendarDivToken = state.push("div_open", "div", 1);
+    mainCalendarDivToken.attrs = [
       ["class", "calendar"],
       ["data-calendar-type", info.attrs.type || "dynamic"],
-      ["data-calendar-default-view", info.attrs.defaultView || "month"]
+      ["data-calendar-default-view", info.attrs.defaultView || "month"],
+      ["data-calendar-default-timezone", info.attrs.tzdefault]
```
jjaffeux: maybe `info.attrs.defaultTimezone` would be more consistent with the data attribute name and what we do for `defautView`

assets/stylesheets/common/discourse-calendar.scss
```diff
@@ -35,6 +36,10 @@
       padding: 3px 3px 3px 0.5em;
     }
 
+    .fc-center {
+      visibility: hidden;
```
jjaffeux: why not display: none ?

assets/javascripts/initializers/discourse-calendar.js.es6
```diff
@@ -298,6 +307,35 @@ function initializeDiscourseCalendar(api) {
...
+  }
+
+  function _setupTimezonePicker(calendar, timeZone) {
+    let $timezonePicker = $(".discourse-calendar-timezone-picker");
+
+    if ($timezonePicker.length) {
+      $timezonePicker.on("change", function(e) {
```
jjaffeux: this is very minor nitpick, but I prefer `event` over `e`, easier to grep for event than e

jjaffeux:
Can we do something about the large margin at the bottom of the calendar? would be great if it could be equivalent to the margin on sides.

martin-brennan:
@jjaffeux I've made the fixes now, I'll merge at my end of day

PR #12, opened 2020-01-13
FEATURE: Add to google calendar link on list view

markvanlan:


markvanlan:
Hooked into `datesRender` callback so that everytime events change, the `Add to calendar` link can be added if in list view.

PR #14, opened 2020-01-20
FEATURE: Fill description for events in calendar link

markvanlan:
Screenshot from 2020-01-20 09-22-39
Screenshot from 2020-01-20 09-22-56

The max description length for Google's description is 8000 characters, but I maxed the excerpt at 1000.

### DIFF
--- a/lib/event_updater.rb
+++ b/lib/event_updater.rb
@@ -30,7 +30,7 @@ module DiscourseCalendar
       html = (doc.try(:to_html) || html).sub(' → ', '')
 
       detail = [
-        PrettyText.excerpt(html, 50, strip_links: true, text_entities: true, keep_emoji_images: true),
+        PrettyText.excerpt(html, 1000, strip_links: true, text_entities: true, keep_emoji_images: true),
         from.iso8601.to_s,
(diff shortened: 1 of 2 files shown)


lib/event_updater.rb
```diff
@@ -30,7 +30,7 @@ def self.update(post)
       html = (doc.try(:to_html) || html).sub(' → ', '')
 
       detail = [
-        PrettyText.excerpt(html, 50, strip_links: true, text_entities: true, keep_emoji_images: true),
+        PrettyText.excerpt(html, 1000, strip_links: true, text_entities: true, keep_emoji_images: true),
```
ZogStriP: That might make the popup on the calendar super huge :(

markvanlan: It does make the popup large (if there is a lot of text).

I thought it was quite nice, because you basically read the post without having to leave the calendar.

I could change this though, so that the popup only takes 100 characters or 50. The excerpt here would still be 1000, but in JS I could take less for the popup.

ZogStriP: I guess we could try with 1000 and see how it feels.

PR #18, opened 2020-01-28
FIX: All-day event logic fixed for events starting at midnight

markvanlan:


markvanlan:
@davidtaylorhq This fixes the issues we were seeing earlier. The problem was that all events starting at midnight were assumed to be all-day even when that was not the case.

I also found a bug in the logic for adjusting for site settings. It is very confusing, but you can see the tests make more sense now.

markvanlan:
@davidtaylorhq Also I set events that have a set start time, but no end time, to be an hour long.

PR #23, opened 2020-03-09
DEV: Migrate from custom fields to tables

nbianca:


ZogStriP:
You could have used a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) instead of adding `(Do not merge)` on the title 😉

jjaffeux:
@nbianca I will request @vinothkannans too, because he did many changes in this lib

db/migrate/20200310200000_remove_timezone_custom_field.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RemoveTimezoneCustomField < ActiveRecord::Migration[5.2]
+  def change
+    UserCustomField.where(name: 'timezone').delete_all
```
eviltrout: Can you use SQL here instead? We've had issues historically with objects in migrations.

db/migrate/20200226183018_create_calendar_events.rb
```diff
@@ -0,0 +1,31 @@
...
+      t.string :region
+      t.timestamps
+
+      t.index :topic_id
+      t.index :post_id
+      t.index :user_id
+    end
+
+    # Rebuild calendar events
+    calendar_topic_ids = Post
```
eviltrout: The following should be SQL for future compatibility.

nbianca: I agree with you here, but building a SQL is not a trivial task because calendar events used to be stored in OP only and now I am moving them in a `CalendarEvent` which belongs to the post where the event was extracted from.

db/migrate/20200310200000_remove_timezone_custom_field.rb
nbianca: Of course, I will do this.

db/migrate/20200226183018_create_calendar_events.rb
eviltrout: The thing is we have to do something, because if you leave a reference to a model in here and the structure of that model ever changes the old migration will break. By putting a reference to a model in a migration means the migration will never run unless that model exists with the exact same structure!

One thing you could do is convert the 1st and 3rd queries to SQL. Then for the second query that loops through posts, surround the whole thing with a rescue `nil`. Normally I would not recommend swallowing errors but in the case that the migration fails it's better to not have any calendar info than break the ability to run migrations at all.

jjaffeux:
I think we should merge this ASAP, I will probably change quite a lot of things with the event update coming as it's kinda making two times the job on few parts.

jjaffeux:
@nbianca tried it on a backup and I have the following error when loading an availability topic:

```
undefined local variable or method `date' for #<PostSerializer:0x00007fd5663baae0>
```

PR #37, opened 2020-04-16
Change height to min-height for group-timezone sizing

markvanlan:
Right now, the larger sizes look really weird with only a few users in a group. This allows the group timezone component to look good with small # of users, but also max at the correct height.


jjaffeux:
This will cause topic height jumping on load no ?

PR #44, opened 2020-05-20
FEATURE: Add user preferences field to select region

udan11:
To determine user's region, it uses moment.js `guess` method and a map between timezones and regions. The list of supported regions is copied from the holidays gem (with the exception of the "pseudo-regions").


assets/javascripts/discourse/components/region-input.js.es6
```diff
@@ -0,0 +1,288 @@
+import ComboBoxComponent from "select-kit/components/combo-box";
+import { computed } from "@ember/object";
+
+// List of regions supported by the 'holidays' gem
+// require 'holidays' ; Holidays.available_regions
+const HOLIDAY_REGIONS = [
```
ZogStriP, line 6: Can we have this list automatically generated by `Holidays.available_regions` minus the ones that are too specific?

assets/javascripts/connectors/user-custom-preferences/region.js.es6
```diff
@@ -0,0 +1,452 @@
+import { action } from "@ember/object";
+
+const TIME_ZONE_TO_REGION = {
```
ZogStriP, line 3: How did you generate this list?

udan11: I generated the list from [another one](https://github.com/moment/moment-timezone/blob/develop/data/meta/latest.json). Technically, I think we could update moment.js to include all the data we need.

assets/javascripts/discourse/components/region-input.js.es6
udan11: I can render this on the server with a `.js.erb` file. I will try that.

ZogStriP: That might not work once we migrate to ember-cli though.

assets/javascripts/connectors/user-custom-preferences/region.js.es6
ZogStriP: At the very least we could have a rake task that would update it?

eviltrout: Yes do not use `.js.erb` - we have a rake task in core `javascript:update_constants` - can a plugin add to that?

udan11: Yes, a plugin can do that. Simply by defining another rake task `javascript:update_constants` will concatenate to the existent one.

assets/javascripts/discourse/components/region-input.js.es6
udan11: I implemented a rake task for it.

assets/javascripts/discourse/components/region-input.js.es6
```diff
@@ -0,0 +1,30 @@
...
+
+  selectKitOptions: {
+    filterable: true,
+    allowAny: false
+  },
+
+  content: computed(function() {
+    const localeNames = {};
+
+    const availableLocales = JSON.parse(this.siteSettings.available_locales);
```
ZogStriP, line 18: Do we need that temp variable?

lib/tasks/javascript.rake
```diff
@@ -0,0 +1,20 @@
+TIMEZONES_DEFINITIONS = 'https://raw.githubusercontent.com/moment/moment-timezone/develop/data/meta/latest.json'
+
+task 'javascript:update_constants' => :environment do
+  require 'holidays'
+  holiday_regions = Holidays.available_regions.map(&:to_s) - ["ecbtarget", "federalreserve", "federalreservebanks", "fedex", "nerc", "unitednations", "ups"]
```
ZogStriP, line 5: I would extract this array in a constant with a helpful name (like `UNUSED_REGIONS`)

PR #51, opened 2020-07-21
Meeting location and google calendar link

ti0:
- Option to add a location for an event, which could additionally have a link
- Display a link to quickly add event to google calendar
- Location gets added to link so it is saved in the google calendar event.

To create a link with location, the format is as follows:
Format: `<Link Text>~<Link url>`
Example: Zoom~https://zoom.us

For location with link, it appears like so:
image

When using text-only location, it appears like so:
image

Without a location, it appears like so:
image


jjaffeux:
I'm sorry thanks for you work, but we are going to implement this differently. I'm going to close this.

PR #58, opened 2020-08-17
HACK: Drop readonly tables when running tests

romanrizzi:
To reproduce:
  - Clone this repo.
  - ` bundle exec rake db:drop db:create db:migrate LOAD_PLUGINS=1 RAILS_ENV=test`
  - `LOAD_PLUGINS=1 bundle exec rspec /Users/romeer/projects/discourse/spec/lib/db_helper_spec.rb`

Fails with this error:

```
Failure/Error:
        rows = DB.exec(<<~SQL, from: from, to: to, like: like)
          UPDATE \"#{table}\"
             SET #{set}
           WHERE #{where}
        SQL

PG::RaiseException:
  ERROR:  Discourse: discourse_calendar_post_events is read only
  CONTEXT:  PL/pgSQL function discourse_functions.raise_discourse_calendar_post_events_readonly() line 3 at RAISE
# (eval):24:in `exec'
# (eval):24:in `async_exec'
```

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -34,6 +34,21 @@ register_svg_icon 'fas fa-star'
 register_svg_icon 'fas fa-file-upload'
 
 after_initialize do
+
+  # TODO: Put this code inside a migration.
+  # lib/db_helper_spec.rb fails without this because these tables are read-only
+  # See: https://github.com/discourse/discourse-calendar/commit/c17bbf57d15920101319b0be40f0bc0fbf17802a
+  if Rails.env.test?
+    connection = ActiveRecord::Base.connection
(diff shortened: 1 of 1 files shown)


jjaffeux:
Make sure you read this too please @tgxworld :

https://meta.discourse.org/t/using-object-storage-for-uploads-s3-clones/148916/125

tgxworld:
Looks like remap doesn't work so well with readonly tables so we can just go ahead and drop the tables. I kept them around previously just in case something goes wrong with the migration. 

https://github.com/discourse/discourse-calendar/commit/f13cd3ebcafa4bfef571eaeba394dcb0bee8416d

PR #60, opened 2020-09-22
PERF: optimizes missing_user_ids query

jjaffeux:


jjaffeux:
Note: I need to dig why the self.raw_invitees check is needed here

app/models/discourse_post_event/event.rb
```diff
@@ -286,12 +286,34 @@ def self.update_from_raw(post)
...
+      missing_users_sql_query = <<-SQL
+        SELECT "group_users"."user_id"
+        FROM "group_users"
+        INNER JOIN "groups"
+          ON "groups"."id" = "group_users"."group_id"
+      SQL
+
+      if self.raw_invitees.present?
+        missing_users_sql_query += <<-SQL
+          AND "groups"."name" IN ('#{self.raw_invitees.join("', '")}')
```
eviltrout, line 299: This is a little scary to me. The combination of the word `raw` with interpolated SQL definitely raises an eyebrow for me.

Does mini sql not allow us to pass an array as a sanitized parameter?

jjaffeux: Yes I'm not happy of this yet, will rework on it

PR #63, opened 2020-10-07
FIX: ensures we have a post

jjaffeux:
This could fail in specific cases like discourse-styleguide where we don't get a real post.

### DIFF
--- a/assets/javascripts/initializers/discourse-post-event-decorator.js.es6
+++ b/assets/javascripts/initializers/discourse-post-event-decorator.js.es6
@@ -195,7 +195,7 @@ function initializeDiscoursePostEventDecorator(api) {
       if (helper) {
         const post = helper.getModel();
 
-        if (post.event) {
+        if (post?.event) {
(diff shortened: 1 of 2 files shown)


assets/javascripts/initializers/discourse-post-event-decorator.js.es6
```diff
@@ -195,7 +195,7 @@ function initializeDiscoursePostEventDecorator(api) {
       if (helper) {
         const post = helper.getModel();
 
-        if (post.event) {
+        if (post?.event) {
```
ZogStriP, line 198: Wow, that's a thing in ES6?

ZogStriP (approved):

assets/javascripts/initializers/discourse-post-event-decorator.js.es6
jjaffeux: sir yes sir, it's part of my work on updating babel

PR #69, opened 2020-11-11
FIX: typo causing syntax error and failed rebuild

md-misko:
This was a real silly mistake
  setting: ""value
but it went through lint anyways.

Perhaps there should be a stricter syntax check for config files, requiring quotes on values and only allowing
  setting:
    setting: "value"


md-misko:
Just noticed this was already caught and fixed just moments before this pull request, so nothing to fix any more.

PR #71, opened 2020-11-19
FIX: sort calendar_details for better holidays support

ZogStriP:
Made sure that events/holidays are loaded in order to prevent an edge case where events would not show up.

Also did a tiny bit of refactoring/cleanup in `create_holiday_events.rb`


jjaffeux (approved):
:+1:

PR #73, opened 2020-11-22
Extracting category id using regex

leog:
Apply the calendar plugin on enabled categories with their default URL different from `/c/category-name/id`.


pmusaraj:
Closing due to inactivity.

PR #78, opened 2020-11-26
FIX: remove ActiveRecord from event data migration

lis2:
Remove as much as possible of ActiveRecord and ruby code from event data migration.

### DIFF
--- a/db/post_migrate/20201111005205_move_data_to_event_dates.rb
+++ b/db/post_migrate/20201111005205_move_data_to_event_dates.rb
@@ -1,41 +1,103 @@
 # frozen_string_literal: true
 
 class MoveDataToEventDates < ActiveRecord::Migration[6.0]
+
+  VALID_OPTIONS = [
+    :start,
+    :end,
+    :status,
+    :"allowed-groups",
+    :url,
+    :name,
+    :reminders,
+    :recurrence
+  ]
+
+  def extract_events(post)
+    cooked = PrettyText.cook(post.raw, topic_id: post.topic_id, user_id: post.user_id)
+    valid_options = VALID_OPTIONS.map { |o| "data-#{o}" }
+
+    valid_custom_fields = []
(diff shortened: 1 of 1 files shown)


db/post_migrate/20201111005205_move_data_to_event_dates.rb
```diff
@@ -1,41 +1,103 @@
...
+    :status,
+    :"allowed-groups",
+    :url,
+    :name,
+    :reminders,
+    :recurrence
+  ]
+
+  def extract_events(post)
+    cooked = PrettyText.cook(post.raw, topic_id: post.topic_id, user_id: post.user_id)
```
SamSaffron, line 17: do you need to re-cook? Maybe just select the cooked column?

db/post_migrate/20201111005205_move_data_to_event_dates.rb
```diff
@@ -1,41 +1,103 @@
...
+    valid_options = VALID_OPTIONS.map { |o| "data-#{o}" }
+
+    valid_custom_fields = []
+    SiteSetting.discourse_post_event_allowed_custom_fields.split('|').each do |setting|
```
SamSaffron, line 21: small one, but this relies on the site setting which may be renamed... instead just select form the table

db/post_migrate/20201111005205_move_data_to_event_dates.rb
```diff
@@ -1,41 +1,103 @@
...
+    :end,
+    :status,
+    :"allowed-groups",
+    :url,
+    :name,
+    :reminders,
+    :recurrence
+  ]
+
+  def extract_events(post)
```
SamSaffron, line 15: minor: a bit odd how you call it with "event" and then it becomes post

SamSaffron (approved):

SamSaffron:
except for the minor comment good to merge, just be sure you test this on local on a few events first.

PR #118, opened 2021-04-13
FIX: Handle multiple holidays in the same day

udan11:
If a user had two holidays in the same day, the calendar would show
only one of them, but the username would be shown twice. This ensures
the calendar shows both holidays and the username only once.


eviltrout (approved):

ZogStriP (approved):
What is time? 🤷

PR #120, opened 2021-04-23
FIX: Show holiday name only once

udan11:
It used to handle the case when the same user has multiple holidays in
the same day by joining all holiday names in a single string. This was
a problem when another user was in the same region and had the same
holidays, because each holiday name was duplicated.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -422,9 +422,9 @@ after_initialize do
     end
 
     grouped.each do |_, v|
-      v[:name] = v[:name].sort.join(", ")
-      v[:usernames].sort!
-      v[:usernames].uniq!
+      v[:name].sort!.uniq!
+      v[:name] = v[:name].join(", ")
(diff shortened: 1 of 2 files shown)


plugin.rb
```diff
@@ -422,9 +422,9 @@ def valid_event
     end
 
     grouped.each do |_, v|
-      v[:name] = v[:name].sort.join(", ")
-      v[:usernames].sort!
-      v[:usernames].uniq!
+      v[:name].sort!.uniq!
```
ZogStriP, line 425: You can chain bang methods?! 😲

udan11: `sort!` yes, because it always returns itself. `uniq!` no, because it returns `nil` if there were no items removed.

ZogStriP: Weird to have different behavior. Usually, you can't chain bang methods.

eviltrout (approved):

PR #121, opened 2021-04-24
UX: Update event notification text and markup

awesomerobot:
Needed an extra span in there to match core styling (default text color username, blue link), also shortened text of the default event notification to match other core notification formatting

Before:
Screen Shot 2021-04-23 at 11 31 09 PM

After:
Screen Shot 2021-04-23 at 11 03 32 PM


eviltrout (approved):
Looks good to me!

PR #122, opened 2021-04-27
UX: makes date time range multiline

jjaffeux:
This is due to new native picker taking more inline space than pikaday.


jjaffeux:
Will now look like this:

Capture d’écran 2021-04-27 à 11 01 22

ZogStriP (approved):

PR #123, opened 2021-04-27
SECURITY: prevents arbitrary method call from input

jjaffeux:


jobs/scheduled/monitor_event_dates.rb
```diff
@@ -46,9 +46,14 @@ def due_reminders(event_date)
       return [] if event_date.event.reminders.blank?
       event_date.event.reminders.split(",").map do |reminder|
         value, unit = reminder.split('.')
-        date = event_date.starts_at - value.to_i.send(unit)
-        { description: reminder, date: date }
-      end.select { |reminder| reminder[:date] <= Time.current }.sort_by { |reminder| reminder[:date] }.drop(event_date.reminder_counter)
+
+        begin
+          date = event_date.starts_at - value.to_i.public_send(unit)
```
udan11, line 55: I do not think there are so many acceptable values for unit (second, minute, hour, day, month and maybe year? plus their plural forms), so what I would do instead is check for those, instead of the begin-rescue dance.

This gives better control over what gets called. In the future, when this will no longer be an experimental feature I guess we will need the units list for validation anyway.

db/post_migrate/20201111005205_move_data_to_event_dates.rb
```diff
@@ -51,7 +51,7 @@ def due_reminders(event)
     return [] if event.reminders.blank?
     event.reminders.split(",").map do |reminder|
       value, unit = reminder.split('.')
-      date = event.original_starts_at - value.to_i.send(unit)
+      date = event.original_starts_at - value.to_i.public_send(unit)
```
udan11, line 54: Malformed events will cause this migration to explode and may stop deployments.

davidtaylorhq (approved):

jobs/scheduled/monitor_event_dates.rb
davidtaylorhq: Oh yes agreed. Now that we have the `validate_reminder_unit` check, the begin/rescue can probably be removed.

> plus their plural forms

Looks like `ActiveSupport::Duration::PARTS` includes only the plural forms. So I guess we need to allow the singular form as well?

```
[1] pry(main)> ActiveSupport::Duration::PARTS
=> [:years, :months, :weeks, :days, :hours, :minutes, :seconds]
```

jjaffeux: No the builder will output `1.minutes` and I think it's fine, at least for now. Maybe we can deal with this later, but any error coming from this is manual input error.

spec/jobs/scheduled/monitor_event_dates_spec.rb
```diff
@@ -97,4 +97,34 @@
       expect(events).to include(event_name: :discourse_post_event_event_ended, params: [past_event])
     end
   end
+
+  context '#due_reminders' do
```
jjaffeux, line 101: this test is half good, as you could just have the fact that 1.foo doesn't make the due_reminders in range... but other alternatives are testing private method which doesn’t ensure the method will be used,

db/post_migrate/20201111005205_move_data_to_event_dates.rb
```diff
@@ -51,9 +51,14 @@ def due_reminders(event)
...
-      date = event.original_starts_at - value.to_i.send(unit)
-      { description: reminder, date: date }
-    end.select { |reminder| reminder[:date] <= Time.current }.sort_by { |reminder| reminder[:date] }
+
+      begin
+        date = event.original_starts_at - value.to_i.public_send(unit)
+        { description: reminder, date: date }
+      rescue NoMethodError
+        nil
+      end
```
davidtaylorhq, line 60: This is still vulnerable to the security issue, right? How about

```suggestion
      allowed = ["years", "months", "weeks", "days", "hours", "minutes", "seconds"]
      next if allowed.include?(unit)
      date = event.original_starts_at - value.to_i.public_send(unit)
      { description: reminder, date: date }
```
(`allowed` could also be moved into a constant at the top of the file)

jjaffeux: it seems like it's very improbable it would happen or do any arm with the public send and the fact that should be with vulnerable data on something that's hasn't been migrated yet... but yes I can also add extra safety here

davidtaylorhq: Agreed it's improbable, but it's not impossible ;)

db/post_migrate/20201111005205_move_data_to_event_dates.rb
```diff
@@ -51,9 +51,12 @@ def due_reminders(event)
     return [] if event.reminders.blank?
     event.reminders.split(",").map do |reminder|
       value, unit = reminder.split('.')
-      date = event.original_starts_at - value.to_i.send(unit)
+
+      allowed = ["years", "months", "weeks", "days", "hours", "minutes", "seconds"]
+      next if allowed.include?(unit)
```
davidtaylorhq, line 56:
```suggestion
      next if !allowed.include?(unit)
```
(my bad, the logic in my suggestion was backwards)

udan11 (approved):

davidtaylorhq (approved):

PR #129, opened 2021-05-03
Fixed typo in german translation

larskn:


CvX:
Hi, we accept changes to translations via https://translate.discourse.org/ platform. Thank you for contributing! 😃

PR #136, opened 2021-05-31
DEV: Allow custom holidays to be defined for the plugin.

tgxworld:


SamSaffron:
hmmm can we fix mistakes as well?

tgxworld:
Nope we can only add new holidays not override or correct existing ones.

PR #138, opened 2021-06-01
Hard fork holidays/holidays gem into plugin.

tgxworld:
The original gem does not seem to be maintained anymore so we're pulling
this into the plugin so that we can update the definitions easily.


eviltrout (approved):
It's impossible to review 376 files, but I can approve the "change" here. I agree with moving the gem into `vendor`

PR #141, opened 2021-06-07
FIX: Independence Day should use to_weekday_if_weekend(date) for all US

tshenry:
When Independence Day does not fall on a weekend, it is observed on a weekday. This was only happening for the state of Virginia. It has been updated to conform with the way things are handled for holidays like New Years Day and Christmas.

Looking at this reference site listed here:

https://github.com/discourse/discourse-calendar/blob/d469260bf99d1d2f29cd32d3904f8c663e102ed0/vendor/holidays/definitions/us.yaml#L66

I'm seeing in the "Previous Years" section that in 2017 Virginia was the only state listed as observing the holiday which explains why it was limited to that state in the code prior to this change. It seems like there has been some arbitrary decision making on a per-state basis, but currently it is nationally observed according to that source.


tgxworld:
Be sure to run `rake generate:definitions` as well after updating the definitions file :) https://github.com/discourse/discourse-calendar#holidays

SamSaffron (approved):

PR #160, opened 2021-07-27
FIX: fix hidden calendar weeks on month view

ryantm:
https://meta.discourse.org/t/discourse-calendar/97376/248?u=ryantm for details

### DIFF
--- a/assets/stylesheets/common/discourse-calendar.scss
+++ b/assets/stylesheets/common/discourse-calendar.scss
@@ -4,7 +4,6 @@
 }
 
 .category-calendar .calendar {
-  height: 400px;
   overflow-y: scroll;
 }
 


jjaffeux:
Sorry but unless Im misunderstanding something here (didnt look at code for a long time) we cant do this or we will have variable height calendars which will lead to topic jumpyness.

ryantm:
> Sorry but unless Im misunderstanding something here (didnt look at code for a long time) we cant do this or we will have variable height calendars which will lead to topic jumpyness.

@jjaffeux Yes, I think that is right. However, currently scrolling is broken, so something needs to be done. I'm not familiar enough with the CSS to fix that, I think.

ryantm:
@jjaffeux I looked at it a bit more, and it looks like https://github.com/discourse/discourse-calendar/pull/160/files#diff-372f425feff0e1a9254d5d4713f92ba53aefb499bf28751450e479cbd3553b75L16 limiting the height to 645px should still apply, so I think this change is okay. It will mean the category topics start lower down, but it will be a fixed height.

jjaffeux:
ok nice, will merge it tomorrow and track how it goes 👍

jjaffeux (approved):

jjaffeux:
merged thanks ❤️

PR #194, opened 2021-12-14
DEV: adds support for event url in event_started trigger

jjaffeux:


lis2 (approved):

tgxworld:
I'm sorry to ask but do we not need tests here?

jjaffeux:
> I'm sorry to ask but do we not need tests here?

Don't be. So the answer is yes, but testing plugins which relies on other plugins is kinda cumberstone ATM. I have various strategies to mitigate this: heavy testing in automation core and I try to test core behavior of script in the plugins (aka not testing automation but that the core of the script is doing what we wanted)

PR #207, opened 2022-01-04
FEATURE: fullDay calendar option

lis2:
Allow calendar to disrespect timezones and always display full day for standalone and group events.

Standalone events are saved in specific timezone: https://github.com/discourse/discourse-calendar/blob/main/app/models/calendar_event.rb#L62

However, we don't store information about timezone. It is essential to have that information to calculate date properly. Therefore, a new column was added.

### DIFF
--- a/assets/javascripts/initializers/discourse-calendar.js.es6
+++ b/assets/javascripts/initializers/discourse-calendar.js.es6
@@ -137,12 +137,13 @@ function initializeDiscourseCalendar(api) {
     const timezone = _getTimeZone($calendar, api.getCurrentUser());
     const calendar = _buildCalendar($calendar, timezone);
     const isStatic = $calendar.attr("data-calendar-type") === "static";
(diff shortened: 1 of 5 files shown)


assets/javascripts/initializers/discourse-calendar.js.es6
```diff
@@ -434,30 +434,43 @@ function initializeDiscourseCalendar(api) {
     calendar.addEvent(event);
   }
 
-  function _setDynamicCalendarEvents(calendar, post, siteSettings) {
+  function _setDynamicCalendarEvents(calendar, $calendar, post, siteSettings) {
```
eviltrout, line 437: Could we avoid using jQuery here by passing in the fullDay as a parameter instead of the jQuery reference?

eviltrout (approved):
Looks good now, thanks!

PR #213, opened 2022-01-24
ensure tables width stays constant on zoom

janzenisaac:
When you zoom in on a browser (safari only) it would manipulate the width of the tables causing the browser to read the table as 'overflown' which would trigger the `expand table` option. Ensure that the table keeps a width of 100% to prevent this from happening. No adverse reactions found in testing new change.


eviltrout (approved):

CvX:
👀 `assets/stylesheets/desktop/.discourse-calendar.scss.un~`

janzenisaac:
_Vim 7.3 contains a new feature persistent undo, that is, undo information won't be lost when quitting Vim and be stored in a file that ends with .un~. **You have set the undofile option**_

PR #217, opened 2022-01-25
Prevent table overflow

janzenisaac:
When you zoom in on a browser (safari only) it would manipulate the width of the tables causing the browser to read the table as 'overflown' which would trigger the `expand table` option. Ensure that the table keeps a width of 100% to prevent this from happening. No adverse reactions found in testing new change.

### DIFF
--- a/assets/stylesheets/desktop/discourse-calendar.scss
+++ b/assets/stylesheets/desktop/discourse-calendar.scss
@@ -1,4 +1,7 @@
 .calendar {
+  table {
+    width: 100%;
+  }
   .fc-list-item-add-to-calendar {
     float: right;
     margin-right: 5px;


janzenisaac:
Is having both `assets/stylesheets/desktop/discourse-calendar.scss~` and `assets/stylesheets/desktop/discourse-calendar.scss` wrong? Not sure why this is happening

CvX:
I'm guessing vim is saving that ~ version as a working copy or something. You should add these: https://github.com/github/gitignore/blob/main/Global/Vim.gitignore to your global gitignore (https://gist.github.com/subfuzion/db7f57fff2fb6998a16c)

CvX (approved):

PR #219, opened 2022-02-09
FIX remove arrow even if there is no whitespace around it

ryantm:
See https://meta.discourse.org/t/discourse-calendar/97376/325?u=ryantm for the issue report.


ryantm:
I haven't actually tested this, but looking over my old posts, I see the old ones are still working fine, so I'm more confident this will fix it. @jjaffeux could you take a look?

jjaffeux:
I will start reviewing this next week.

jjaffeux (approved):

jjaffeux:
merged it, thanks ❤️

PR #226, opened 2022-02-22
FIX: enforces the value of recurrence

jjaffeux:
Due to the fact we do this in :post_process_cooked it's easier to ensure no invalid value is set rather than showing some error in the UI as it will be too late.

### DIFF
--- a/app/models/discourse_post_event/event.rb
+++ b/app/models/discourse_post_event/event.rb
@@ -10,6 +10,8 @@ module DiscoursePostEvent
 
     has_many :event_dates, dependent: :destroy
 
+    before_validation :enforce_recurrence
+
     def self.attributes_protected_by_default
       super - %w[id]
     end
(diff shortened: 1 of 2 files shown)


app/models/discourse_post_event/event.rb
```diff
@@ -365,5 +367,21 @@ def calculate_next_date
...
+        every_month
+        every_week
+        every_two_weeks
+        every_day
+        every_weekday
+      ]
+
+      if self.recurrence.present? && !valid_recurrences.include?(self.recurrence)
+        self.recurrence = 'every_week'
+      end
```
nattsw, line 384: > Due to the fact we do this in :post_process_cooked

`post_process_cooked` -> `update_from_raw(post)` -> ...

Do you mean https://github.com/discourse/discourse-calendar/blob/b7e0441e77b64d1230faa9a72deda0c138d80ff3/app/models/discourse_post_event/event.rb#L333-L367?

Does this mean we have duplication?

lib/discourse_post_event/event_validator.rb
```diff
@@ -61,6 +61,20 @@ def validate_event
...
       end
 
+      if extracted_event[:recurrence].present?
+        valid_recurrences = %w[
+          every_month
+          every_week
+          every_two_weeks
+          every_day
+          every_weekday
+        ]
```
CvX, line 71: Can this be extracted into a const?

jjaffeux: done in https://github.com/discourse/discourse-calendar/pull/226/commits/11ddb1160ad9ce57bba4994f3ce2040b2cb7e32d

CvX (approved):

PR #231, opened 2022-02-23
DEV: Remove old ics code

CvX:


lis2 (approved):

cedk:
The ICS file was very useful to import the calendar in an external software.
Would you reconsider included it back?

CvX:
The ics support is still there! 😃 This PR just removed the old implementation. Most of the new ics code lives in core now, see e.g. https://github.com/discourse/discourse/blob/771dddb711b97c10baf9ac5fd23d769f06f38e7e/app/assets/javascripts/discourse/app/lib/plugin-api.js#L1517-L1532

cedk:
But for me https://discuss.tryton.org/discourse-post-event/events.ics is no more working. What is the new URL path for the new implementation?

JulienPalard:
Same as @cedk, I don't feel the need for "one shot ICS export of one event" but I feel the need to subscribe to all events via a single endpoint, so I can subscribe to it from my calendar, like the old `events.ics`.

PR #233, opened 2022-02-24
FIX: apply latest holidays definitions

ZogStriP:
@jordanvidrine needed to run `rake generate:definitions` in the `vendor/holidays` directory to update the *.rb files with the latest holidays definitions 😉


jordanvidrine (approved):
Thanks :+1:

PR #234, opened 2022-02-25
FEATURE: adds a currently away report

jjaffeux:


plugin.rb
```diff
@@ -657,4 +657,25 @@ class Jobs::ExportCsvFile
 
   register_notification_consolidation_plan(reminders_consolidation_plan)
   register_notification_consolidation_plan(invitation_consolidation_plan)
+
+  if Report.respond_to?(:add_report)
```
CvX, line 661: Is it possible for `Report` to not respond to `:add_report`?

jjaffeux: Yes, for some time this API was not present, I guess I could stop doing this check at some point, but that doesn't look urgent and can probably wait a little bit more?

CvX: This API? https://github.com/discourse/discourse/blame/main/app/models/report.rb#L203-L205

CvX (approved):

PR #239, opened 2022-03-03
FEATURE: adds holiday flair to chat

jjaffeux:
Also slightly refactors the code:
- uses decorateCookedElement
- relies on the fact that cancel accepts undefined
- uses optional chaining


jjaffeux:
Screenshot 2022-03-03 at 09 55 52

nbianca (approved):

PR #248, opened 2022-04-11
FEATURE: add events calendar on the category page

techAPJ:
This commit adds a new site setting `events calendar category` for events calendar that is separate and independent from `calendar categories`. If a category is present in both `events calendar category` and `calendar categories` then priority is given to `calendar categories`.

Screen Shot 2022-04-08 at 09 46 42

Screen Shot 2022-04-08 at 09 56 28

### DIFF
--- a/assets/javascripts/initializers/discourse-calendar.js
+++ b/assets/javascripts/initializers/discourse-calendar.js
@@ -41,19 +41,23 @@ function initializeDiscourseCalendar(api) {
   }
 
   api.onPageChange((url) => {
-    const $calendarContainer = $(`${selector}.category-calendar`);
-    if ($calendarContainer.length) {
-      $calendarContainer.hide();
+    const categoryCalendarNode = document.querySelector(
+      `${selector}.category-calendar`
+    );
+    if (categoryCalendarNode) {
+      categoryCalendarNode.innerHTML = "";
     }
 
-    const calendarNode = document.getElementById("upcoming-events-calendar");
-    if (calendarNode) {
-      calendarNode.innerHTML = "";
+    const categoryEventNode = document.getElementById(
+      "upcoming-events-calendar"
+    );
+    if (categoryEventNode) {
+      categoryEventNode.innerHTML = "";
     }
 
     const browsedCategory = Category.findBySlugPathWithID(url);
     if (browsedCategory) {
-      if (!$calendarContainer.length) {
+      if (!categoryCalendarNode) {
         return;
       }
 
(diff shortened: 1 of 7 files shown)


assets/javascripts/initializers/discourse-calendar.js
```diff
@@ -36,14 +39,21 @@ function initializeDiscourseCalendar(api) {
 
   api.onPageChange((url) => {
     const $calendarContainer = $(`${selector}.category-calendar`);
-    if (!$calendarContainer.length) {
-      return;
+    if ($calendarContainer.length) {
+      $calendarContainer.hide();
```
pmusaraj, line 43: I know you didn't add this, but we should be moving away from jQuery. Maybe a first step here can be to avoid using jQuery's hide/show. (Maybe add/remove the `.hidden` class instead?)

Also, if we could strip out some jquery, that would be great too.

assets/javascripts/templates/connectors/discovery-list-container-top/category-events.hbs
```diff
@@ -0,0 +1 @@
+<div id="upcoming-events-calendar"></div>
```
pmusaraj, line 1: Could we use the existing `upcoming-events-calendar` component here? It feels like we would be reusing some code/logic and just need to change the model.

pmusaraj (commented):
Thanks, looks good, a small comment on jQuery and a question on using the existing component. 

This also needs tests, both on the backend and on the frontend.

assets/javascripts/templates/connectors/discovery-list-container-top/category-events.hbs
techAPJ: I tried but the issue here is I am not sure how to set the model from inside the initializer condition.

https://github.com/discourse/discourse-calendar/blob/aff34c012a51f9d6c946de6c3741eebfb5a57a68/assets/javascripts/discourse/templates/discourse-post-event-upcoming-events-index.hbs#L2

pmusaraj: Ah, yes, in the connector it's a little trickier to set this up. It's probably fine as is then.

config/settings.yml
```diff
@@ -72,3 +72,7 @@ discourse_post_event:
     list_type: simple
     client: true
     default: ""
+  events_calendar_category:
```
pmusaraj, line 75: Small thing I noticed here, we should rename this to `events_calendar_categories` because this allows multiple selections (which is good, it's nice that multiple categories can each have their calendar.)

pmusaraj (approved):
Awesome! 

Just curious, how did you solve the tests timeout in CI?

CvX:
https://github.com/discourse/discourse/pull/16533 and https://github.com/discourse/discourse/pull/16536 😃

PR #265, opened 2022-05-10
DEV: Generate holiday definitions from yaml

davidtaylorhq:
followup to ae8b9630f620a53ecc466914c1bc7a72c26de963


CvX (approved):
Oops, forgot about it too 😃

PR #267, opened 2022-05-13
FIX: Make region visible to current user too

nbianca:
The 'holidays-region' custom field used to be visible only to staff
members, but editable by the current user. This meant that the user
could edit the value, but could not see the value after it was set.
This caused the input field from their user preferences page to be
always blank for regular users.

### DIFF
--- a/spec/serializers/user_serializer_spec.rb
+++ b/spec/serializers/user_serializer_spec.rb
@@ -12,7 +12,7 @@ describe UserSerializer do
     user.upsert_custom_fields(DiscourseCalendar::REGION_CUSTOM_FIELD => "uk")
   end
 
-  context "as user" do
+  context "as another user" do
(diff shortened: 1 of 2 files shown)


spec/serializers/user_serializer_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  fab!(:user) { Fabricate(:user) }
+
+  subject { described_class.new(user, scope: guardian).as_json }
+
+  before do
+    SiteSetting.calendar_enabled = true
+    user.upsert_custom_fields(DiscourseCalendar::REGION_CUSTOM_FIELD => "uk")
+  end
+
+  context "as user" do
```
ZogStriP, line 15: I would clarify we're testing for another user.

```suggestion
  context "as another user" do
```

ZogStriP (approved):

PR #269, opened 2022-05-15
DEV: Add API endpoints for holidays

yaki-imo:
This will make two new endpoints available to admins:
* `/admin/discourse-calendar/holiday-regions` - this will return a list of holiday regions (as defined by the Holidays gem)
* `/admin/discourse-calendar/holiday-regions/:id/holidays` - this will return all of the holidays for a given region

The next piece I will work on for this will be a UI update so that admins can select a region and view all of the holidays for that region (and eventually the ability to enable/disable holidays).

I wanted to keep the PRs smaller to make them easier to review, and also to make sure I'm on the right track. Still new to contributing so any feedback is welcome! :smiley:

### DIFF
--- a/app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
+++ b/app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require "holidays"
+
+module Admin::DiscourseCalendar
+  class AdminHolidaysController < AdminDiscourseCalendarController
+    def index
+      region_code = params[:region_code]
+
+      begin
+        holidays = Holidays.year_holidays([region_code], Time.current.beginning_of_year)
+      rescue Holidays::InvalidRegion
+        return render_json_error(I18n.t("system_messages.discourse_calendar_holiday_region_invalid"), 422)
+      end
+
+      render json: { region_code: region_code, holidays: holidays }
+    end
+  end
+end
--- a/plugin.rb
+++ b/plugin.rb
@@ -84,6 +84,19 @@ after_initialize do
     end
   end
 
+  # DISCOURSE CALENDAR
+
+  %w[
+    ../app/controllers/admin/admin_discourse_calendar_controller.rb
+    ../app/controllers/admin/discourse_calendar/admin_holiday_regions_controller.rb
+  ].each { |path| load File.expand_path(path, __FILE__) }
+
+  Discourse::Application.routes.append do
+    mount ::DiscourseCalendar::Engine, at: '/'
+
+    get '/admin/discourse-calendar/holiday-regions' => 'admin/discourse_calendar/admin_holiday_regions#index', constraints: StaffConstraint.new
+  end
+
   # DISCOURSE POST EVENT
 
   %w[
--- a/plugin.rb
+++ b/plugin.rb
@@ -89,12 +89,14 @@ after_initialize do
   %w[
     ../app/controllers/admin/admin_discourse_calendar_controller.rb
     ../app/controllers/admin/discourse_calendar/admin_holiday_regions_controller.rb
+    ../app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
   ].each { |path| load File.expand_path(path, __FILE__) }
 
   Discourse::Application.routes.append do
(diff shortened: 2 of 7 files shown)


plugin.rb
```diff
@@ -84,6 +84,26 @@ class Engine < ::Rails::Engine
...
 
+  # DISCOURSE CALENDAR
+
+  %w[
+    ../app/controllers/admin/admin_discourse_calendar_controller.rb
+    ../app/controllers/admin/discourse_calendar/admin_holiday_regions_controller.rb
+    ../app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
+  ].each { |path| load File.expand_path(path, __FILE__) }
+
+  ::ActionController::Base.prepend_view_path File.expand_path(
```
pmusaraj, line 95: We don't need to prepend the views given that there aren't any yet. And since this is an admin endpoint, I suspect there won't be any Rails views, the Ember frontend will handle this.

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+require "holidays"
+
+module Admin::DiscourseCalendar
+  class AdminHolidaysController < AdminDiscourseCalendarController
+    def index
+      begin
+        holidays = Holidays.year_holidays([params[:id]], Time.current.beginning_of_year)
+      rescue Holidays::InvalidRegion
+        raise Discourse::NotFound
```
pmusaraj, line 11: We should output an `invalid region` error here instead of a 404.

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+      end
+
+      render json: { holidays: holidays }
```
pmusaraj, line 14: Not necessary in this PR, but we might want to include the region name label here, the UI will likely need it. (Same for the holiday regions controller.)

pmusaraj (approved):
Cool, looking good, thanks! Left a few comments, all minor.

plugin.rb
yaki-imo: Oh yes, good catch, thank you! Totally makes sense! :facepalm: :joy: :+1:

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
yaki-imo: Sounds good! :+1: Something like?: 
```
return render_json_error(I18n.t("system_messages.discourse_calendar_holiday_region_invalid"), 422)
```

Also, is `422` what you had in mind? :thinking: :smiley:

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
yaki-imo: Ah yes, of course! An oversight on my part, I suppose it would have surfaced at some point! :sweat_smile: :smiley: I can take care of it here, something like this for the return payload?:
```
{ region: params[:id], holidays: holidays }
```

Or thoughts on `region_code` or `region_id`? :thinking:

pmusaraj: I was thinking it would be useful to have both the region code (`ar`) and a pretty name to display (like `Holidays in Argentina`) but it doesn't look like the gem has that. So might be best to leave this for later, we'll probably want to just use localizations based on the region code.

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
yaki-imo: I just updated this to return a 422 and an error message. :smiley:

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
yaki-imo: > I was thinking it would be useful to have both the region code (ar) and a pretty name to display (like Holidays in Argentina) but it doesn't look like the gem has that. 

Ah, makes sense! Well, to start I updated this to return the region code at least. :smiley:

pmusaraj (approved):

PR #270, opened 2022-05-16
FIX: various TZ-related issues

ZogStriP:
This is a WIP to fix issues in discourse-events, notably when handling timezones, but also various UI quirks.

I'm leaving it here rather than on my HDD while I'm on holiday in case anyone wants to pick it up.

### DIFF
--- a/app/models/discourse_post_event/event.rb
+++ b/app/models/discourse_post_event/event.rb
@@ -4,43 +4,58 @@ module DiscoursePostEvent
   class Event < ActiveRecord::Base
     PUBLIC_GROUP = 'trust_level_0'
 
-    self.table_name = 'discourse_post_event_events'
+    MIN_NAME_LENGTH = 5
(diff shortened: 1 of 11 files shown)


cloud-construct:
Is this to fix where an event is exported to an ICS file or Google Calendar (it creates the event +1 on London time)

ZogStriP:
> Is this to fix where an event is exported to an ICS file or Google Calendar (it creates the event +1 on London time)

I haven't tested that part. But will make sure it's fixed in the process as well 👍

cloud-construct:
> > Is this to fix where an event is exported to an ICS file or Google Calendar (it creates the event +1 on London time)
> 
> I haven't tested that part. But will make sure it's fixed in the process as well 👍

That'd be awesome, thanks.

davidtaylorhq:
Continued in #299

PR #283, opened 2022-06-14
Allow admins to enable and disable holidays

yaki-imo:
The functionality in this PR will allow admins to enable and disable holidays.

Some notes about disabled holidays:
- They won't get added to the calendar
- For holidays already on the calendar, they will get removed when they are disabled

An overview of the changes in this PR:
- Disabled holidays will get saved in a new `disabled_holidays` table
- A new `Holiday` service was added that will find holidays (from the Holidays gem) and attach a `disabled` field
- Two backend endpoints were added to disable and enable holidays
- A new `admin-holidays-list-item` component was added that will display an individual holiday and a disable or enable button, this component also includes the AJAX calls to hit the new corresponding backend endpoints

All feedback welcome! :+1: :smiley:

### DIFF
--- a/app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
+++ b/app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
@@ -1,19 +1,41 @@
 # frozen_string_literal: true
 
-require "holidays"
-
 module Admin::DiscourseCalendar
   class AdminHolidaysController < AdminDiscourseCalendarController
     def index
       region_code = params[:region_code]
 
       begin
-        holidays = Holidays.year_holidays([region_code], Time.current.beginning_of_year)
+        holidays = DiscourseCalendar::Holiday.find_holidays_for(region_code: region_code)
       rescue Holidays::InvalidRegion
         return render_json_error(I18n.t("system_messages.discourse_calendar_holiday_region_invalid"), 422)
       end
 
       render json: { region_code: region_code, holidays: holidays }
+    end
+
+    def disable
+      DiscourseCalendar::DisabledHoliday.create!(disabled_holiday_params)
+      CalendarEvent.destroy_by(
+        description: disabled_holiday_params[:holiday_name],
+        region: disabled_holiday_params[:region_code]
+      )
+
+      render json: success_json
+    end
+
+    def enable
+      if DiscourseCalendar::DisabledHoliday.destroy_by(disabled_holiday_params).present?
+        render json: success_json
+      else
+        render_json_error(I18n.t("system_messages.discourse_calendar_enable_holiday_failed"), 422)
+      end
+    end
+
+    private
+
+    def disabled_holiday_params
+      params.require(:disabled_holiday).permit(:holiday_name, :region_code)
     end
   end
 end
--- a/app/models/discourse_calendar/disabled_holiday.rb
+++ b/app/models/discourse_calendar/disabled_holiday.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module DiscourseCalendar
+  class DisabledHoliday < ActiveRecord::Base
+    validates :holiday_name, presence: true
+    validates :region_code, presence: true
+  end
+end
+
+# == Schema Information
+#
+# Table name: discourse_calendar_disabled_holidays
+#
+#  id           :bigint           not null, primary key
+#  holiday_name :string           not null
+#  region_code  :string           not null
+#  disabled     :boolean          default(TRUE), not null
+#  created_at   :datetime         not null
+#  updated_at   :datetime         not null
+#
+# Indexes
+#
+#  index_disabled_holidays_on_holiday_name_and_region_code  (holiday_name,region_code)
+#
--- a/app/services/discourse_calendar/holiday.rb
+++ b/app/services/discourse_calendar/holiday.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require "holidays"
+
+module DiscourseCalendar
+  class Holiday
+    def self.find_holidays_for(
+      region_code:,
+      start_date: Date.current.beginning_of_year,
+      end_date: Date.current.end_of_year,
+      show_holiday_observed_on_dates: false)
+
+      holidays = Holidays.between(
+        start_date,
+        end_date,
+        [region_code],
+        show_holiday_observed_on_dates ? :observed : [])
+
+      holidays.map do |holiday|
+        holiday[:disabled] = DiscourseCalendar::DisabledHoliday
+          .where(region_code: region_code)
+          .exists?(holiday_name: holiday[:name])
+      end
+
+      holidays
+    end
+  end
+end
--- a/assets/javascripts/discourse/templates/components/admin-holidays-list-item.hbs
+++ b/assets/javascripts/discourse/templates/components/admin-holidays-list-item.hbs
@@ -1,0 +1,15 @@
+<td>{{holiday.date}}</td>
+<td>{{holiday.name}}</td>
+<td>
+  {{#if isHolidayDisabled}}
+    {{d-button
+      action=(action "enableHoliday" holiday region_code)
+      label="discourse_calendar.enable_holiday"
+    }}
+  {{else}}
+    {{d-button
+      action=(action "disableHoliday" holiday region_code)
+      label="discourse_calendar.disable_holiday"
+    }}
+  {{/if}}
+</td>
--- a/jobs/scheduled/create_holiday_events.rb
+++ b/jobs/scheduled/create_holiday_events.rb
@@ -40,9 +40,13 @@ module Jobs
         .delete_all
 
       regions_and_user_ids.each do |region, user_ids|
-        Holidays
-          .between(Date.today, 6.months.from_now, [region], :observed)
-          .filter { |holiday| (1..5) === holiday[:date].wday }
+        DiscourseCalendar::Holiday.find_holidays_for(
+          region_code: region,
+          start_date: Date.today,
+          end_date: 6.months.from_now,
+          show_holiday_observed_on_dates: true
+        )
+          .filter { |holiday| (1..5) === holiday[:date].wday && holiday[:disabled] === false }
           .each do |holiday|
 
           user_ids.each do |user_id|
--- a/plugin.rb
+++ b/plugin.rb
@@ -85,13 +85,15 @@ after_initialize do
     end
   end
 
-  # DISCOURSE CALENDAR
+  # DISCOURSE CALENDAR HOLIDAYS
 
   add_admin_route 'admin.calendar', 'calendar'
 
   %w[
     ../app/controllers/admin/admin_discourse_calendar_controller.rb
     ../app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
+    ../app/models/discourse_calendar/disabled_holiday.rb
+    ../app/services/discourse_calendar/holiday.rb
   ].each { |path| load File.expand_path(path, __FILE__) }
 
   Discourse::Application.routes.append do
@@ -99,6 +101,8 @@ after_initialize do
 
     get '/admin/plugins/calendar' => 'admin/plugins#index', constraints: StaffConstraint.new
     get '/admin/discourse-calendar/holiday-regions/:region_code/holidays' => 'admin/discourse_calendar/admin_holidays#index', constraints: StaffConstraint.new
+    post '/admin/discourse-calendar/holidays/disable' => 'admin/discourse_calendar/admin_holidays#disable', constraints: StaffConstraint.new
+    delete '/admin/discourse-calendar/holidays/enable' => 'admin/discourse_calendar/admin_holidays#enable', constraints: StaffConstraint.new
   end
 
   # DISCOURSE POST EVENT
--- a/spec/jobs/scheduled/create_holiday_events_spec.rb
+++ b/spec/jobs/scheduled/create_holiday_events_spec.rb
@@ -84,6 +84,37 @@ describe DiscourseCalendar::CreateHolidayEvents do
     expect(CalendarEvent.exists?(username: frenchy.username)).to eq(false)
   end
 
+  context "when there are disabled holidays" do
+    let(:france_assomption) { { holiday_name: "Assomption", region_code: "fr" } }
+    let(:france_toussaint) { { holiday_name: "Toussaint", region_code: "fr" } }
+
+    before do
+      DiscourseCalendar::DisabledHoliday.create!(france_assomption)
+      DiscourseCalendar::DisabledHoliday.create!(france_toussaint)
+    end
+
+    it "only adds enabled holidays to the calendar" do
+      frenchy
+      freeze_time Time.zone.local(2019, 7, 1)
+      subject.execute(nil)
+
+      expect(CalendarEvent.pluck(:region, :description, :start_date, :username)).to eq([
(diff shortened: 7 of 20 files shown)


app/services/discourse_calendar/holiday.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require "holidays"
+
+module DiscourseCalendar
+  class Holiday
```
yaki-imo, line 6: My thought behind this new holiday service is for it to be an interface and wrapper for the Holidays gem. It will return the results from the gem and also add a new `disabled` field. I also created it so I didn't have to duplicate the disabled functionality - the admin UI needs this functionality so it knows which holidays are enable/disabled, and the `create_holiday_events` job needs it to only add enabled holidays to the calendar. I also think this is where custom holidays would get incorporated at.

spec/jobs/scheduled/create_holiday_events_spec.rb
```diff
@@ -84,6 +84,37 @@
     expect(CalendarEvent.exists?(username: frenchy.username)).to eq(false)
   end
 
+  context "when there are disabled holidays" do
```
yaki-imo, line 87: I added these tests (a lot of it copied from existing tests here), but for some reason I can't get them to pass locally. I keep getting "Discourse::InvalidAccess: can_create? failed" when I run the specs via `bundle exec rake autospec`, despite this, it looks like they're passing the build though. :raised_eyebrow: :confused:

plugin.rb
```diff
@@ -85,20 +85,24 @@ class Engine < ::Rails::Engine
...
+    ../app/services/discourse_calendar/holiday.rb
   ].each { |path| load File.expand_path(path, __FILE__) }
 
   Discourse::Application.routes.append do
     mount ::DiscourseCalendar::Engine, at: '/'
 
     get '/admin/plugins/calendar' => 'admin/plugins#index', constraints: StaffConstraint.new
     get '/admin/discourse-calendar/holiday-regions/:region_code/holidays' => 'admin/discourse_calendar/admin_holidays#index', constraints: StaffConstraint.new
+    post '/admin/discourse-calendar/holidays/disable' => 'admin/discourse_calendar/admin_holidays#disable', constraints: StaffConstraint.new
+    delete '/admin/discourse-calendar/holidays/enable' => 'admin/discourse_calendar/admin_holidays#enable', constraints: StaffConstraint.new
```
pmusaraj, line 105: It feels a little odd to use `delete` and append `/enable` to the URL. Maybe we can do this instead: 

```suggestion
    delete '/admin/discourse-calendar/holidays/disable' => 'admin/discourse_calendar/admin_holidays#enable', constraints: StaffConstraint.new
```

That also feels a bit odd, to be honest.

spec/jobs/scheduled/create_holiday_events_spec.rb
pmusaraj: You're probably just missing `LOAD_PLUGINS=1` when running autospec. I pulled the branch locally and `LOAD_PLUGINS=1 bin/rspec plugins/discourse-calendar/spec` works fine. 👍

spec/requests/admin/admin_holidays_controller_spec.rb
```diff
@@ -65,5 +65,102 @@ module Admin::DiscourseCalendar
...
+          before do
+            sign_in(admin)
+          end
+
+          it "disables a specific holiday in a specific region" do
+            post "/admin/discourse-calendar/holidays/disable.json",
+              params: {
+                disabled_holiday: dia_de_la_independencia
+              }
+
```
romanrizzi, line 84:
```suggestion
disabled_holiday = DiscourseCalendar::DisabledHoliday.last
```

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
```diff
@@ -1,19 +1,39 @@
...
+        holidays = DiscourseCalendar::Holiday.find_holidays_for(region_code: region_code)
       rescue Holidays::InvalidRegion
         return render_json_error(I18n.t("system_messages.discourse_calendar_holiday_region_invalid"), 422)
       end
 
       render json: { region_code: region_code, holidays: holidays }
     end
+
+    def disable
+      DiscourseCalendar::DisabledHoliday.create(disabled_holiday_params)
```
romanrizzi, line 18: We should validate the parameters before saving them in case there's a bug in the client.

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
```diff
@@ -1,19 +1,39 @@
...
+      CalendarEvent.delete_by(
```
romanrizzi, line 19:
```suggestion
      CalendarEvent.destroy_by(
```

`#delete_by` doesn't run callbacks. Should we check what happens if there's no `CalendarEvent`?

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
```diff
@@ -1,19 +1,39 @@
...
+        description: disabled_holiday_params[:holiday_name],
+        region: disabled_holiday_params[:region_code]
+      )
+
+      render json: success_json
+    end
+
+    def enable
+      DiscourseCalendar::DisabledHoliday.delete_by(disabled_holiday_params)
```
romanrizzi, line 28:
```suggestion
      DiscourseCalendar::DisabledHoliday.destroy_by(disabled_holiday_params)
```

Should we also handle the case where there's no record?

jobs/scheduled/create_holiday_events.rb
```diff
@@ -40,9 +40,13 @@ def execute(args)
...
-        Holidays
-          .between(Date.today, 6.months.from_now, [region], :observed)
-          .filter { |holiday| (1..5) === holiday[:date].wday }
+        DiscourseCalendar::Holiday.find_holidays_for(
+          region_code: region,
+          start_date: Date.today,
+          end_date: 6.months.from_now,
+          show_holidays_observed_on_dates: true
+        )
+          .filter { |holiday| (1..5) === holiday[:date].wday && holiday[:disabled] === false }
```
romanrizzi, line 49: Should we move this filter inside the service in case we want to re-use it in the future?

spec/requests/admin/admin_holidays_controller_spec.rb
```diff
@@ -65,5 +65,102 @@ module Admin::DiscourseCalendar
...
+        context "when an admin is signed in" do
+          before do
+            sign_in(admin)
+          end
+
+          context "when there is a disabled holiday" do
+            let(:hong_kong_labour_day) { { holiday_name: "Labour Day", region_code: "hk" } }
+
+            before do
+              DiscourseCalendar::DisabledHoliday.create(hong_kong_labour_day)
```
romanrizzi, line 148: As a rule of thumb, we prefer the `!` version when calling create, save, or update.

spec/requests/admin/admin_holidays_controller_spec.rb
```diff
@@ -65,5 +65,102 @@ module Admin::DiscourseCalendar
...
+              }
+
+            expect(DiscourseCalendar::DisabledHoliday.last.disabled).to eq(true)
+            expect(DiscourseCalendar::DisabledHoliday.last.holiday_name).to eq(dia_de_la_independencia[:holiday_name])
+            expect(DiscourseCalendar::DisabledHoliday.last.region_code).to eq(dia_de_la_independencia[:region_code])
+          end
+
+          context "when a holiday has been added to the calendar" do
+            let(:calendar_post) { create_post(raw: "[calendar]\n[/calendar]") }
+            let(:australia_new_years_day) { { holiday_name: "New Year's Day", date: "2022-01-01", region_code: "au" } }
```
romanrizzi, line 92:
```suggestion
            let(:australia_new_years_day) do
              {  
                holiday_name: "New Year's Day", 
                date: "2022-01-01", 
                region_code: "au" 
              }
            end
```

app/models/discourse_calendar/disabled_holiday.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+module DiscourseCalendar
+  class DisabledHoliday < ActiveRecord::Base
```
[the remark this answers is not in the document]
yaki-imo: Thanks for the tip! :+1:  I added the annotations here, and for two model below that were missing them.

assets/javascripts/discourse/templates/components/admin-holidays-list-item.hbs
```diff
@@ -0,0 +1,19 @@
+<tr class={{if isHolidayDisabled "disabled" "enabled"}}>
+  <td>{{holiday.date}}</td>
+  <td>{{holiday.name}}</td>
+  <td>
+    {{#if isHolidayDisabled}}
+      {{d-button
+        action=(action "enableHoliday" holiday region_code)
+        ariaLabel="discourse_calendar.enable_holiday"
```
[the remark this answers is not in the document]
yaki-imo: Thanks for pointing this out, I removed the unnecessary aria label.

app/services/discourse_calendar/holiday.rb
```diff
@@ -0,0 +1,30 @@
...
+        end_date,
+        [region_code],
+        show_holidays_observed_on_dates ? :observed : [])
+
+      disabled_holiday_names = DiscourseCalendar::DisabledHoliday
+        .where(region_code: region_code)
+        .pluck(:holiday_name)
+
+      holidays.map do |holiday|
+        holiday[:disabled] = disabled_holiday_names.include?(holiday[:name]) ? true : false
```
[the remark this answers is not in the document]
yaki-imo: Excellent suggestion! I updated it! :+1:

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
yaki-imo: I added some Rails `presence` validations for `holiday_name` and `region_code`, I also updated the migration to make sure they're not null.

spec/jobs/scheduled/create_holiday_events_spec.rb
yaki-imo: Yup, that was it! I ran that command and it worked locally! Thanks!

spec/requests/admin/admin_holidays_controller_spec.rb
yaki-imo: Nice suggestion, I made this change! :+1:

spec/requests/admin/admin_holidays_controller_spec.rb
yaki-imo: I made this change, I also made a few other lines shorter by wrapping them! :+1:

spec/requests/admin/admin_holidays_controller_spec.rb
yaki-imo: Thanks for mentioning this! I updated this and the other ones I added to use the `!` version.

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
yaki-imo: I updated this to use `destroy_by`. I added a check to see if a record was deleted, if one was deleted it'll return a 200 and if one wasn't it'll return a 422. I added some tests for these cases too.

assets/javascripts/discourse/templates/components/admin-holidays-list-item.hbs
```diff
@@ -0,0 +1,19 @@
+<tr class={{if isHolidayDisabled "disabled" "enabled"}}>
```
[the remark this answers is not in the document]
yaki-imo: Thanks for this! I added it in! Re: `enabled`, I added that in to make the tests read better, but you're right, it doesn't need to be there, so I removed it and updated the tests accordingly. :+1:

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
yaki-imo: I also added a test case for when the parameters are missing.

romanrizzi: That's great! Should we also check for invalid region codes? I assume the holiday name could be anything, or should we validate that against the gem holidays too?

yaki-imo:
> I think it would be good to add some explanatory text in the initial admin screen, below the region selector:
>
> Something along the lines of
> 
> > Pick a region to see the scheduled holidays for that region.
> > Any disabled holidays will be automatically excluded from the staff holiday calendar.

Great suggestion! I added some explanatory text, just tweaked this text :point_up: slightly! :smile:

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
yaki-imo: I updated this to use `destroy_by`. I was thinking about checking if a calendar event was deleted or not, but I can't seem to find a reason to check for that. One thought came to me though, maybe if the UI alerted a message if an event was deleted or not. Wondering if you had something in mind for this? :thinking: :smile:

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
yaki-imo: I don't think it'll be too difficult to validate the region code against the Holidays gem, I can work on that. :+1:  Maybe we could even do the holiday names too, I'll check that out too!

plugin.rb
yaki-imo: Not bad! I think that's better than what I had because it's deleting the `disabled_holiday` record. I can make this change! :+1:

spec/services/discouse-calendar/holiday_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+        ))
+      end
+
+      describe "dates holidays are observed on" do
+        let(:holidays) { Holiday.find_holidays_for(
+          region_code: "sg",
+          start_date: "2021-12-31",
+          end_date: "2022-05-31",
+          show_holiday_observed_on_dates: show_holiday_observed_on_dates) }
+
```
yaki-imo, line 41: I added the test cases below for returning the dates holidays are observed on. Just a random FYI about this (in case it's not clear/known), the observed on dates get added to the calendar, but the admin shows the actual holiday dates, not the observed on dates. Which is why I tweaked the explanatory text to remove "scheduled" (as in "scheduled holidays"), because I didn't want it to cause confusion that the dates appearing in the admin wouldn't necessarily be the same dates that would get removed from the calendar, if that makes sense! :sweat_smile: 

And another side note, at first glance I don't think the Holidays gem can return both dates, but it could be done by making two calls to the gem - if we think it would be useful to display both dates in the admin. :smile:

pmusaraj:
@romanrizzi the changes here look good to me, if you're OK with them as well, can you approve/merge next week? I was going to do that, but ran out of time this week, and am off next week. Thanks!

app/controllers/admin/discourse_calendar/admin_holidays_controller.rb
romanrizzi: Looks great now. Thanks!

romanrizzi:
PR looks great. Thank you very much, @yaki-imo!

yaki-imo:
> PR looks great. Thank you very much, @yaki-imo!

:+1: No problem! :smile:

PR #284, opened 2022-06-15
FIX: Optimize performance of UnescapeEventName migration on large sites

tgxworld:
On large sites, the migration can take up to 4 hours which is not
something we want. This commit adds the following optimisations:

1. Add an index for notifications `WHERE notifcation_type IN (27, 28)`
1. Increase each batch size to 10000, the smaller the batch size the more
times we have to scan the notifications table.
1. Avoid having to start the scan of the notifications table from id 1
and instead start from the first event notification type.

### DIFF
--- a/db/post_migrate/20220613073844_unescape_event_name.rb
+++ b/db/post_migrate/20220613073844_unescape_event_name.rb
@@ -3,10 +3,14 @@
 class UnescapeEventName < ActiveRecord::Migration[7.0]
   disable_ddl_transaction!
 
+  TEMP_INDEX_NAME = "_temp_discourse_calendar_unescape_event_name_migration"
+
   def up
     # event notifications
-    start = 1
-    limit = DB.query_single("SELECT MAX(id) FROM notifications WHERE notification_type IN (27, 28)").first.to_i
+    DB.exec("CREATE INDEX CONCURRENTLY #{TEMP_INDEX_NAME} ON notifications(id) WHERE notification_type IN (27, 28)")
+    start, limit = DB.query_single("SELECT MIN(id), MAX(id) FROM notifications WHERE notification_type IN (27, 28)")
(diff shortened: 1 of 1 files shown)


nattsw (approved):

nattsw:
> Increase each batch size to 1000

I suppose you mean 10000?

tgxworld:
> I suppose you mean 10000?

oops yea I meant 10000

db/post_migrate/20220613073844_unescape_event_name.rb
```diff
@@ -3,10 +3,14 @@
 class UnescapeEventName < ActiveRecord::Migration[7.0]
   disable_ddl_transaction!
 
+  TEMP_INDEX_NAME = "_temp_discourse_calendar_unescape_event_name_migration"
+
   def up
     # event notifications
-    start = 1
-    limit = DB.query_single("SELECT MAX(id) FROM notifications WHERE notification_type IN (27, 28)").first.to_i
+    DB.exec("CREATE INDEX CONCURRENTLY #{TEMP_INDEX_NAME} ON notifications(id) WHERE notification_type IN (27, 28)")
```
martin-brennan, line 10: Should this have `UnescapeEventName::` prefix too like the drop?

martin-brennan: Also why `ON notifications(id)` instead of just `ON notifications`?

tgxworld: Oops nope, not necessary to have the prefix.

martin-brennan: Oh wait NVM I see!

martin-brennan (approved):
Nice!

PR #299, opened 2022-07-24
FIX: Handle non-UTC events correctly

davidtaylorhq:
Previously we were storing a 'UTC' datetime in the database which was in fact a local datetime. That means that, every time it was accessed, consumers would have to ignore the "Z" suffix and parse it using the specified `timezone`.

Adding that kind of logic to every consumer would be very error-prone. Instead, this commit changes the implementation so that `original_start_time` and `original_end_time` are 'technically correct', truly-UTC timestamps. Consumers that need to obtain the original 'local' timestamp (e.g. recurrence) can do something like `original_start_time.in_timezone(timezone)`. Most consumers (e.g. notifications, event-countdown UI, etc.) don't care about the original timezone, and so they can now stick to using `original_start_time` without issue.

A number of specs are introduced to test timezone behaviour, including recurrence across daylight-saving transitions.

A migration will fix existing erroneous data in the database.

A number of JS fixes are also included, and a basic JS acceptance test is introduced.

### DIFF
--- a/db/migrate/20220724130519_fix_post_event_timezones.rb
+++ b/db/migrate/20220724130519_fix_post_event_timezones.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+class FixPostEventTimezones < ActiveRecord::Migration[7.0]
+  def up
+    execute <<~SQL
+      UPDATE discourse_post_event_events
+      SET
+        original_starts_at = (original_starts_at::timestamp AT TIME ZONE timezone),
+        original_ends_at = (original_ends_at::timestamp AT TIME ZONE timezone)
+      WHERE timezone IS NOT NULL;
+    SQL
+
+    execute <<~SQL
+      UPDATE discourse_calendar_post_event_dates
+      SET
+        starts_at = (starts_at::timestamp AT TIME ZONE timezone),
+        ends_at = (ends_at::timestamp AT TIME ZONE timezone)
+      FROM discourse_post_event_events
+      WHERE discourse_post_event_events.id = discourse_calendar_post_event_dates.event_id
+      AND discourse_post_event_events.timezone IS NOT NULL
+    SQL
+
+    execute <<~SQL
+      UPDATE topic_custom_fields
+      SET value = (value::timestamp AT TIME ZONE discourse_post_event_events.timezone) AT TIME ZONE 'UTC'
+      FROM discourse_post_event_events
+      JOIN posts ON discourse_post_event_events.id = posts.id
+      WHERE discourse_post_event_events.timezone IS NOT NULL
+      AND topic_custom_fields.topic_id = posts.topic_id
+      AND topic_custom_fields.name IN ('TopicEventStartsAt', 'TopicEventEndsAt')
+    SQL
+  end
+end
--- a/spec/acceptance/post_spec.rb
+++ b/spec/acceptance/post_spec.rb
@@ -579,16 +579,16 @@ describe Post do
     it "stores the correct information in the database" do
       expected_datetime = ActiveSupport::TimeZone["Australia/Sydney"].parse("2022-07-24 14:01")
 
-      p = PostCreator.create!(
+      post = PostCreator.create!(
         user,
         title: 'Beach party',
         raw: "[event start='2022-07-24 14:01' timezone='Australia/Sydney']\n[/event]"
       ).reload
 
-      expect(p.event.timezone).to eq("Australia/Sydney")
-      expect(p.event.original_starts_at).to eq_time(expected_datetime)
-      expect(p.event.starts_at).to eq_time(expected_datetime)
-      expect(p.event.event_dates.first.starts_at).to eq_time(expected_datetime)
+      expect(post.event.timezone).to eq("Australia/Sydney")
+      expect(post.event.original_starts_at).to eq_time(expected_datetime)
+      expect(post.event.starts_at).to eq_time(expected_datetime)
+      expect(post.event.event_dates.first.starts_at).to eq_time(expected_datetime)
     end
 
     it "raises an error for invalid timezone" do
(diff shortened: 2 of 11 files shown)


spec/acceptance/post_spec.rb
```diff
@@ -570,4 +570,68 @@
...
+      p = PostCreator.create!(
+        user,
+        title: 'Beach party',
+        raw: "[event start='2022-07-24 14:01' timezone='Australia/Sydney']\n[/event]"
+      ).reload
+
+      expect(p.event.timezone).to eq("Australia/Sydney")
+      expect(p.event.original_starts_at).to eq_time(expected_datetime)
+      expect(p.event.starts_at).to eq_time(expected_datetime)
+      expect(p.event.event_dates.first.starts_at).to eq_time(expected_datetime)
```
tgxworld, line 591:
```suggestion
      post = PostCreator.create!(
        user,
        title: 'Beach party',
        raw: "[event start='2022-07-24 14:01' timezone='Australia/Sydney']\n[/event]"
      ).reload

      expect(post.event.timezone).to eq("Australia/Sydney")
      expect(post.event.original_starts_at).to eq_time(expected_datetime)
      expect(post.event.starts_at).to eq_time(expected_datetime)
      expect(post.event.event_dates.first.starts_at).to eq_time(expected_datetime)
```

IMO, the characters we save typing here are not worth the impact on readability.

test/javascripts/acceptance/post-event-builder-test.js
```diff
@@ -0,0 +1,69 @@
...
+    discourse_post_event_allowed_custom_fields: "",
+  });
+
+  test("composer event builder", async function (assert) {
+    await visit("/");
+    await click("#create-topic");
+    await click(".toolbar-popup-menu-options .dropdown-select-box-header");
+    await click(".toolbar-popup-menu-options *[data-value='insertEvent']");
+
+    const m = ".discourse-post-event-builder-modal";
```
tgxworld, line 22: Minor but same comments as https://github.com/discourse/discourse-calendar/pull/299/files#r928366796. There is very little upside in abbreviating variable names compared to its impact on readability.

db/migrate/20220724130519_fix_post_event_timezones.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class FixPostEventTimezones < ActiveRecord::Migration[7.0]
+  def up
+    execute <<~SQL
+      UPDATE discourse_post_event_events
+      SET
+        original_starts_at = (original_starts_at::timestamp AT TIME ZONE timezone),
+        original_ends_at = (original_ends_at::timestamp AT TIME ZONE timezone)
+      WHERE timezone IS NOT NULL;
```
tgxworld, line 10: Is the `timezone` stored always UTC here? I'm a little confused here based on the description of the PR. 

> so that original_start_time and original_end_time are 'technically correct', truly-UTC timestamps.

What I understood from the above is that we'll always store the timestamps as UTC and consumers can figure out the timezone specific time by relying on the `timezone` column.

tgxworld: Just to confirm my understanding, historically `original_starts_at` and `original_ends_at` have always been timestamps based on the `timezone` column and what we're trying to do here is to convert the timestamps to UTC?

davidtaylorhq: An example might be useful here. Let's say we have an event at midday in London currently (UTC +1).

Previously, we were storing it in the database as 

```
original_starts_at: 2022-07-25T12:00Z
timezone: Europe/London
```
So if you take the `original_starts_at` value in isolation, it will be one hour out. Reading the timestamp correctly would require ignoring the `Z` value and forcing `Europe/London` on the parser.

Now, we store the same event as:

```
original_starts_at: 2022-07-25T11:00Z
timezone: Europe/London
```
That means that `original_starts_at` is correct in isolation. 2022-07-025T11:00Z is technically equal to midday in London. For most consumers, that's all they need to know. If the origin timezone is required, then there are easy ruby/js APIs to translate the time into Europe/London. For example:

```
pry(main)> DateTime.parse("2022-07-25T11:00Z").in_time_zone("Europe/London")
=> Mon, 25 Jul 2022 12:00:00.000000000 BST +01:00
```

---

Followup question might be: why don't we store it as `2022-07-25T12:00+01:00` in the database? Postgres stores all timestamps internally as UTC. You can technically pass a localised timestamp to `UPDATE` or `INSERT`, but it will be immediately translated to UTC for storage.

tgxworld: > Now, we store the same event as:
> 
> ```
> original_starts_at: 2022-07-25T11:00Z
> timezone: Europe/London
> ```

But based on the implementation, won't we end up storing `original_starts_at` as 13:00 instead? 

```
discourse_development=# SELECT '2022-07-25T12:00Z' AT TIME ZONE 'Europe/London';
      timezone       
---------------------
 2022-07-25 13:00:00
(1 row)
```

davidtaylorhq: > `'2022-07-25T12:00Z' AT TIME ZONE 'Europe/London';`

We strip the `Z` using `::timestamp`, so that then it is converted to 11:00:00Z

```
# SELECT '2022-07-25T12:00Z'::timestamp AT TIME ZONE 'Europe/London';
        timezone        
------------------------
 2022-07-25 11:00:00+00
(1 row)
```

This kind of hack is only used for the migration of invalid data. Going forward we won't need to do things like this.

tgxworld: Ahhh OK I missed the `::timestamp` part. Thank you for clarifying :+1:

tgxworld (approved):
Changes look good to me :+1:

PR #300, opened 2022-07-25
FIX: consistent desert island icon for holidays

lis2:
In some places like mention and assign, we displayed calendar-alt for holiday. It should be consistent, and we should display tropical island everywhere.

Before:
Screen Shot 2022-07-25 at 1 50 04 pm
Screen Shot 2022-07-25 at 2 10 57 pm

After
Screen Shot 2022-07-25 at 1 49 43 pm
Screen Shot 2022-07-25 at 2 15 55 pm

### DIFF
--- a/test/javascripts/acceptance/holiday-flair-test.js
+++ b/test/javascripts/acceptance/holiday-flair-test.js
@@ -73,6 +73,9 @@ acceptance("Discourse Calendar - Holiday Flair", function (needs) {
   test("shows holiday emoji on mention", async function (assert) {
     await visit("/t/1-3-0beta9-no-rate-limit-popups/28830");
     assert.ok(exists(".mention.on-holiday img.on-holiday"));
-    assert.strictEqual(query(".mention.on-holiday").innerText.trim(), "@eviltrout");
+    assert.strictEqual(
+      query(".mention.on-holiday").innerText.trim(),
+      "@eviltrout"
+    );
   });
 });
--- a/test/javascripts/acceptance/holiday-flair-test.js
+++ b/test/javascripts/acceptance/holiday-flair-test.js
@@ -73,6 +73,6 @@ acceptance("Discourse Calendar - Holiday Flair", function (needs) {
   test("shows holiday emoji on mention", async function (assert) {
     await visit("/t/1-3-0beta9-no-rate-limit-popups/28830");
     assert.ok(exists(".mention.on-holiday img.on-holiday"));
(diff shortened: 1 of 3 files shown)


tgxworld:
Is there a reason we skipped adding a test here?

martin-brennan:
@tgxworld is there any reason to add a test for a correct emoji? Seems pretty unimportant IMO

martin-brennan (approved):

tgxworld:
> is there any reason to add a test for a correct emoji? Seems pretty unimportant IMO

The test isn't to test that the emoji is correct but that the emoji is displayed as intended. I don't think we even have a test for that now. I would have expected a test to break when we changed from an icon to an emoji.

lis2:
> I don't think we even have a test for that now. I would have expected a test to break when we changed from an icon to an emoji.

Sounds fair, let me add one to cover it

lis2:
Test will fail until this one is merged https://github.com/discourse/discourse/pull/17660

```javascript
  const href = getURL(`/u/${username.toLowerCase()}`);
  const mentions = element.querySelectorAll(`a.mention[href="${href}"]`);

  mentions.forEach((mention) => {
```

tgxworld:
> const href = getURL(`/u/${username.toLowerCase()}`);

Where does this come from? I think it needs to support both `/u` and `/users`.

tgxworld:
> Test will fail until this one is merged https://github.com/discourse/discourse/pull/17660

I've merged Test will fail until this one is merged https://github.com/discourse/discourse/pull/17660

test/javascripts/acceptance/holiday-flair-test.js
```diff
@@ -65,4 +69,10 @@ acceptance("Discourse Calendar - Holiday Flair", function (needs) {
     assert.ok(exists("div[data-username='foo'] .holiday-flair"));
     assert.ok(!exists("div[data-username='bar'] .holiday-flair"));
   });
+
+  test("shows holiday emoji on mention", async function (assert) {
+    await visit("/t/1-3-0beta9-no-rate-limit-popups/28830");
+    assert.ok(exists(".mention.on-holiday img.on-holiday"));
+    assert.equal(query(".mention.on-holiday").innerText.trim(), "@eviltrout");
```
tgxworld, line 76:
```suggestion
    assert.strictEqual(query(".mention.on-holiday").innerText.trim(), "@eviltrout");
```

tgxworld (approved):
Looks good to me :+1: Thank you for adding the test.

PR #307, opened 2022-07-29
Update event.rb

nathan-nz:
Matching the event name length to the default topic title length in Discourse core. If there is no event name entered, defaults to topic title anyway.

Has been requested several times, 30 character limit is extremely restrictive.

It would be even better if this used the discourse_max_topic_title_length setting instead to ensure consistency where instances have changed this setting.


SamSaffron:
@jjaffeux I am accepting this for now, long term we want to couple to the site setting and also have some testing here.

SamSaffron (approved):

PR #310, opened 2022-08-15
DEV: Update service references

keegangeorge:
**Update** the following services references:
- `site-settings:main` ->  `service:site-settings`
- `site-main` -> `service:site`

to remove the following deprecation warnings:
```bash
[PLUGIN discourse-calendar] Deprecation notice: "site-settings:main" is deprecated, use "service:site-settings" instead (deprecated since Discourse 2.9.0.beta7) (removal in Discourse 3.0.0)

[PLUGIN discourse-calendar] Deprecation notice: "site:main" is deprecated, use "service:site" instead (deprecated since Discourse 2.9.0.beta7) (removal in Discourse 3.0.0)
```

### DIFF
--- a/assets/javascripts/initializers/add-event-ui-builder.js
+++ b/assets/javascripts/initializers/add-event-ui-builder.js
@@ -55,7 +55,7 @@ export default {
   name: "add-discourse-post-event-builder",
 
   initialize(container) {
-    const siteSettings = container.lookup("site-settings:main");
+    const siteSettings = container.lookup("service:site-settings");
     if (siteSettings.discourse_post_event_enabled) {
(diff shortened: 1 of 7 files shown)


davidtaylorhq (approved):
LGTM 👌 

In general, we need to be careful with this kind of change in themes/plugins because they need to work on the stable branch of Discourse. However, this plugin has [already been configured](https://github.com/discourse/discourse-calendar/blob/e9c99c8be93dce411169e6d4c2e26dfc43180682/.discourse-compatibility#L1) to pin itself to an older commit when installed on Discourse stable, so that concern doesn't apply 🚀

PR #312, opened 2022-08-18
FIX: Update regions translation to include Ghana

s3lase:
Ensure new Ghana holiday region is properly translated and displayed in UI


pmusaraj (approved):
Cool! (Relaunched the test suite, although this shouldn't impact it.)

PR #313, opened 2022-08-19
DEV: Add support for the experimental user menu

OsamaSayegh:
This PR ports the widgets-based implementation of the `event_reminder` and `event_invitation` notification items to the experimental Glimmer-based user menu. The changes in this PR should make event notifications in the experimental user menu consistent with they look/behave in the old user menu.

We don't need to add a `.discourse-compatibility` entry for this change because it's fully backward compatible since we check that the `registerNotificationTypeRenderer` plugin API is available before we attempt to use it.

For more context on the experimental user menu, see https://github.com/discourse/discourse/pull/17379.

### DIFF
--- a/assets/javascripts/discourse/widgets/event-invitation-notification-item.js
+++ b/assets/javascripts/discourse/widgets/event-invitation-notification-item.js
@@ -4,7 +4,7 @@ import { DefaultNotificationItem } from "discourse/widgets/default-notification-
 import { escapeExpression, formatUsername } from "discourse/lib/utilities";
 import { iconNode } from "discourse-common/lib/icon-library";
 
-// TODO: delete the strings marked with TODO in translation files when
+// TODO(osama): delete the strings marked with TODO in translation files when
 // this file is removed
 
 createWidgetFrom(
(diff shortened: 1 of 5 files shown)


assets/javascripts/discourse/widgets/event-invitation-notification-item.js
```diff
@@ -4,6 +4,9 @@ import { DefaultNotificationItem } from "discourse/widgets/default-notification-
 import { escapeExpression, formatUsername } from "discourse/lib/utilities";
 import { iconNode } from "discourse-common/lib/icon-library";
 
+// TODO: delete the strings marked with TODO in translation files when
```
martin-brennan, line 7: This is nothing to do with this PR specifically, but I often find it easier to find my own TODOs in the future if i add the suffix (name) e.g. `# TODO (martin)`. Don't have to follow this just thought I would note it :)

martin-brennan (approved):
LGTM

assets/javascripts/discourse/widgets/event-invitation-notification-item.js
OsamaSayegh, line 7:
```suggestion
// TODO(osama): delete the strings marked with TODO in translation files when
```

OsamaSayegh: done!

PR #315, opened 2022-08-31
FIX: Reintroduce holiday sub-regions removed due to consolidation

s3lase:
US holiday definition previously defined with `us, us_az, us_co, us_oh, us_or, us_sd, us_wy` regions got consolidated down to just `us`. This removed them from the available holiday regions list resulting in errors for users with these regions

This change makes these regions available once again and ensures the list is sorted to avoid noisy diffs generated by constantly changing order of the regions.


Flink (approved):

s3lase:
Thanks @Flink

PR #334, opened 2022-10-18
FIX: Disabling of header sorting now checks the correct site setting and defaults to false

frank3manuel:
- Disabling of header sorting now checks the correct site setting and defaults to false
- Corresponding qunit test fixed to set the category's disable_topic_resorting custom_field before running the test

### DIFF
--- a/test/javascripts/acceptance/sort-event-topics-test.js
+++ b/test/javascripts/acceptance/sort-event-topics-test.js
@@ -1,18 +1,20 @@
 import { acceptance, exists } from "discourse/tests/helpers/qunit-helpers";
 import { test } from "qunit";
 import { visit } from "@ember/test-helpers";
-import { cloneJSON } from "discourse-common/lib/object";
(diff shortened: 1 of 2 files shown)


test/javascripts/acceptance/sort-event-topics-test.js
```diff
@@ -1,18 +1,20 @@
...
-    server.get("/c/1/show.json", () => helper.response(categoryResponse));
+  needs.settings({
+    calendar_enabled: true,
+    discourse_post_event_enabled: true,
+    disable_resorting_on_categories_enabled: true,
   });
 
   test("visiting a category page", async function (assert) {
+    const site = Site.current();
+    site.categories[15].custom_fields = { disable_topic_resorting: true };
```
CvX, line 16: This would have leaked into other tests but I fixed that in https://github.com/discourse/discourse/pull/18663 😄

CvX (approved):

PR #342, opened 2022-11-03
FEATURE: handle subsequent and intersected holidays

AndrewPrigorshnev:
If a user has several holidays one after another, we want to show the farthest end date.

For example, let's say today is Monday, and I am sick, and I also have days off from Tuesday to Friday. Then we want to show Friday as an end date of my time off.

Or let's say I have a vacation from Monday to Friday and at the same time there is a public holiday in my country on Monday. We want to show Friday as the end of my time off in this case.

### DIFF
--- a/lib/users_on_holiday.rb
+++ b/lib/users_on_holiday.rb
@@ -15,14 +15,14 @@ module DiscourseCalendar
     private
 
     def self.current_holiday(user_events)
-      holiday_ends_at = holiday_ends_at(user_events)
-      return nil unless holiday_ends_at
+      ends_at = holiday_ends_at(user_events)
+      return nil unless ends_at
 
       [
         user_events[0].user_id,
         {
           username: user_events[0].username,
-          ends_at: holiday_ends_at
+          ends_at: ends_at
         }
       ]
     end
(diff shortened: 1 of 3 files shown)


lib/users_on_holiday.rb
```diff
@@ -5,17 +5,72 @@ class UsersOnHoliday
...
-        ]
+        .map { |_, events| current_holiday(events) }
+        .compact
+        .to_h
+    end
+
+    private
+
+    def self.current_holiday(user_events)
+      holiday_ends_at = holiday_ends_at(user_events)
```
ZogStriP, line 18: I usually tend to avoid naming variables with the same name as functions. How about this instead?

```suggestion
      ends_at = holiday_ends_at(user_events)
```

lib/users_on_holiday.rb
```diff
@@ -5,17 +5,72 @@ class UsersOnHoliday
...
+      return nil if sorted_events.first.in_future?
+      return sorted_events.first.ends_at if events.count === 1
+
+      result = sorted_events.first.ends_at
+      sorted_events.each_cons(2) do |pair|
+        if pair[0].ends_at < pair[1].start_date
+          return result
+        else
+          result = pair[1].ends_at if pair[1].ends_at > result
+        end
```
ZogStriP, line 70: Slightly shorter version?

```suggestion
        if pair[0].ends_at < pair[1].start_date
          return result
        elsif pair[1].ends_at > result
          result = pair[1].ends_at
        end
```

ZogStriP (approved):

AndrewPrigorshnev:
I've applied all suggestions, @ZogStriP thank you for the review!

ZogStriP (approved):

PR #344, opened 2022-11-14
FEATURE: Add option to bump topic

jancernik:
- At the moment, the topic timer doesn't get created correctly. Probably because I'm triggering it the wrong way.
- I'm reusing the CSS reminder classes for the new input. I have to change that in the future.


jancernik:
**TODOS**
- [ ] Limit the amount of "auto-bump topic" reminders from the UI to one
- [ ] Make `topic-timer-info` update when updating the event from the UI (it only updates when changing the event through raw now)
- [ ] Hide `topic-timer-info` edit buttons when the timer is from an event (not sure it's possible)
- [ ] Add tests
- [ ] Add documentation

jobs/scheduled/monitor_event_dates.rb
```diff
@@ -39,14 +39,21 @@ def finish(event_date)
...
-        value, unit = reminder.split('.')
+      event_date.event.reminders.split(',').map do |reminder|
 
-        next if !validate_reminder_unit(unit)
+        reminder_array = reminder.split('.')
+        if reminder_array.length() === 3
+          type, value, unit = reminder_array
+        else
+          value, unit = reminder_array
+        end
```
pmusaraj, line 54: We're doing this on multiple places and it feels a little flimsy. It might be good to move this logic to a helper in the model and call that. Or... add a migration and avoid this altogether. 

Or... since we can only set one bump per topic, the data for that could be on its own column? So in UI, we can still show the dropdown, but in backend we store it separate, i.e.: 

```
reminders: "15.minutes,-30.minutes",
topicBump: "10.minutes",
```

I'm mostly thinking out loud here, happy to hear your thoughts.

jobs/scheduled/monitor_event_dates.rb
```diff
@@ -39,14 +39,21 @@ def finish(event_date)
 
       return if event_date.event.recurrence.blank?
       event_date.event.set_next_date
+      event_date.event.set_topic_bump
```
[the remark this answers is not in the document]
pmusaraj: Actually, on second thought, this is probably useful for recurring events.

pmusaraj (commented):
Nice work, I have a few questions about structure but otherwise it looks good. 

Also, we need a few tests before merging, some of the other items in your checklist can be handled in separate PRs.

jobs/scheduled/monitor_event_dates.rb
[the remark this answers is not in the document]
jancernik: Yes, I did it that way for recurring events. Every minute checks if the event has ended. If it did it calculates the next date and resets the topic timer.

jobs/scheduled/monitor_event_dates.rb
jancernik: > Or... since we can only set one bump per topic, the data for that could be on its own column?

It would be similar to how I made it previously. I wanted to be as consistent as possible across UI, raw, and models. At least for me, it's clearer that way. I think adding a migration for that would be good, though.

pmusaraj: Yeah, I think you're right, consistency is best. Let's look into a migration, that's probably a good next step.

jobs/regular/discourse_post_event/bump_topic.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module Jobs
+  class DiscoursePostEventBumpTopic < ::Jobs::Base
+    sidekiq_options retry: false
+
+    def execute(args)
+      raise Discourse::InvalidParameters.new(:topic_id) if args[:topic_id].blank?
+      raise Discourse::InvalidParameters.new(:date) if args[:date].blank?
+
+      puts "run job"
```
romanrizzi, line 11:
```suggestion
```

jobs/regular/discourse_post_event/bump_topic.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Jobs
+  class DiscoursePostEventBumpTopic < ::Jobs::Base
+    sidekiq_options retry: false
+
+    def execute(args)
+      raise Discourse::InvalidParameters.new(:topic_id) if args[:topic_id].blank?
```
romanrizzi, line 8:
```suggestion
      return if Topic.find_by(id: args[:topic_id])
```

We don't handle these exceptions anywhere, so it's better to just do an early return.

jobs/regular/discourse_post_event/bump_topic.rb
```diff
@@ -0,0 +1,19 @@
...
+      raise Discourse::InvalidParameters.new(:date) if args[:date].blank?
```
romanrizzi, line 9:
```suggestion
      return if args[:date].blank?
```

jobs/regular/discourse_post_event/bump_topic.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      puts "run job"
+      topic = Topic.find_by(id: args[:topic_id].to_i)
+      event_user = User.find_by(id: topic.user_id)
```
romanrizzi, line 13: Could `topic.user_id` or the `#find_by` return nil here?

db/migrate/20221121165352_add_type_field_to_events_reminders.rb
```diff
@@ -0,0 +1,26 @@
...
+      refactored_reminders = []
+      event.reminders.split(',') do |reminder|
+        refactored_reminders.push(reminder.prepend("notification."))
+      end
+      event.reminders = refactored_reminders.join(',')
+      event.save
+    end
+  end
+
+  def down
```
romanrizzi, line 15: I don't think this is necessary, we should raise `ActiveRecord::IrreversibleMigration` here.

db/migrate/20221121165352_add_type_field_to_events_reminders.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class AddTypeFieldToEventsReminders < ActiveRecord::Migration[7.0]
+  def up
+    DiscoursePostEvent::Event.where.not(reminders: nil).find_each do |event|
```
romanrizzi, line 5: As a rule of thumb, we don't use AR for migrations. We need to rewrite this query using `mini_sql`. See [here](https://github.com/discourse/discourse/blob/main/plugins/chat/db/post_migrate/20221117142910_delete_orphaned_channels.rb) for an example.

db/migrate/20221121165352_add_type_field_to_events_reminders.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class AddTypeFieldToEventsReminders < ActiveRecord::Migration[7.0]
```
romanrizzi, line 3: We should move this file to the `post_migrate` folder.

romanrizzi (changes requested):

romanrizzi (changes requested):

app/models/discourse_post_event/event.rb
```diff
@@ -57,6 +57,19 @@ def set_next_date
       notify_missing_invitees!
     end
 
+    def set_topic_bump
+      return if self.post.event.reminders.blank?
```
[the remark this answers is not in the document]
jancernik: You're right! I think I did that for testing something

jobs/regular/discourse_post_event/bump_topic.rb
jancernik: Yep, I saw that one earlier

app/models/discourse_post_event/event.rb
```diff
@@ -57,6 +57,19 @@ def set_next_date
...
+      self.post.event.reminders.split(',').map do |reminder|
+        type, value, unit = reminder.split('.')
+
+        next if type != 'bumpTopic' || !ActiveSupport::Duration::PARTS.any? { |part| part.to_s == unit }
+
+        date = self.post.event.starts_at - value.to_i.public_send(unit)
+        ::Jobs.enqueue(:discourse_post_event_bump_topic, topic_id: self.post.topic_id, date: date)
+        break
```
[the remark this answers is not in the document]
jancernik: Since the topic timer only supports one action per topic, it only takes the first "bumpTopic" reminder

jobs/regular/discourse_post_event/bump_topic.rb
jancernik: I think so, but that case is already handled inside `set_or_create_timer`

app/models/discourse_post_event/event.rb
```diff
@@ -57,6 +57,25 @@ def set_next_date
...
+        reminders.split(',').each do |reminder|
+          type, value, unit = reminder.split('.')
+          next if type != 'bumpTopic' || !validate_reminder_unit(unit)
+
+          date = starts_at - value.to_i.public_send(unit)
+          break
+        end
+      end
+
+      Jobs.enqueue(:discourse_post_event_bump_topic, topic_id: self.post.topic_id, date: date)
```
pmusaraj, line 72: If an event doesn't have a reminder configured, will this job be invoked with `date = nil`?

spec/jobs/regular/discourse_post_event/bump_topic_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      before do
+        Fabricate(:topic_timer,
+        status_type: TopicTimer.types[:bump],
+        execute_at: 3.days.from_now,
+        topic: topic_1,
+        user: admin_1
+       )
+      end
+
+      it 'deletes the timer if type is bump' do
```
pmusaraj, line 53: I'm not sure I understand this use case... should the topic timer get deleted if the date parameter is missing? 

Is this the use case for when an event is edited and a topic bump has been removed?

pmusaraj (commented):
This looks quite close now, I also pulled the branch locally and tested it and it worked as expected. (I have a few minor questions only.) 

@romanrizzi have your comments been addressed?

romanrizzi:
@pmusaraj - Yep, looks good to me 👍

spec/jobs/regular/discourse_post_event/bump_topic_spec.rb
jancernik: This one is not very intuitive, but it's similar to how the [`set_or_create_timer`](https://github.com/discourse/discourse/blob/99fa10eee60dac8cadc54455f2c6a5724c217e34/app/models/topic.rb#L1387)  method and the requests from [`edit-topic-timer.js`](https://github.com/discourse/discourse/blob/6969e9da7e980659b0b6569129ce32fad61fd058/app/assets/javascripts/discourse/app/controllers/edit-topic-timer.js#L228) work. If the time or date isn't set, it will delete the timer.

> Is this the use case for when an event is edited and a topic bump has been removed?

Yes, if you remove the auto-bump topic reminder, it will delete the timer that way.

app/models/discourse_post_event/event.rb
jancernik: Yes, is not super ideal. Maybe we can remove the job altogether in the future

pmusaraj (approved):
Looks great, thanks!

PR #348, opened 2022-11-21
FIX: exclude deleted events from MonitorEventJob

lis2:
When the post is deleted, the linked event is automatically deleted as well. 
Job should not try to process deleted events.


pmusaraj (approved):
Thanks, merging!

PR #352, opened 2022-11-26
FIX: handle localised automatic group names in event validation

angusmcleod:
Context: https://meta.discourse.org/t/event-calendar-plugin-official-broken-publik-events/246731

### DIFF
--- a/lib/discourse_post_event/event_validator.rb
+++ b/lib/discourse_post_event/event_validator.rb
@@ -93,7 +93,11 @@ module DiscoursePostEvent
       return true unless event[:"allowed-groups"]
 
       event[:"allowed-groups"].split(',').each do |group_name|
-        group = Group.find_by(name: group_name)
+        group = begin
+          Group.lookup_group(group_name.to_sym)
+        rescue ArgumentError
+          nil
+        end
 
         if !group || !guardian.can_see_group?(group)
           @post.errors.add(:base, I18n.t("discourse_post_event.errors.models.event.invalid_allowed_groups"))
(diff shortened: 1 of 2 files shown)


lib/discourse_post_event/event_validator.rb
```diff
@@ -93,7 +93,11 @@ def can_invite_groups?(event)
       return true unless event[:"allowed-groups"]
 
       event[:"allowed-groups"].split(',').each do |group_name|
-        group = Group.find_by(name: group_name)
+        group = begin
+          Group.lookup_group(group_name.to_sym)
```
SamSaffron, line 97: this api in core is a bit awkward, we should just do the simbolizing in core and keep the api as a string api... nothing to change here... but it is certainly odd.

lib/discourse_post_event/event_validator.rb
```diff
@@ -93,7 +93,11 @@ def can_invite_groups?(event)
...
+        rescue ArgumentError
+          nil
```
SamSaffron, line 99: this argument error is odd... how can that happen?

lib/discourse_post_event/event_validator.rb
angusmcleod: Yeah I thought the same.

lib/discourse_post_event/event_validator.rb
angusmcleod: If there's no automatic or stored group matching the argument (the name) it'll be throw [here](https://github.com/discourse/discourse/blob/main/app/models/group.rb#L616). Rather than throw an error here we want `nil` because then the subsequent conditional will fail returning the right error.

SamSaffron (approved):

PR #357, opened 2022-12-20
FIX: Delete topic timer when reminder is removed

jancernik:
Now, if you remove an "auto-bump topic" reminder, it will also remove the topic timer


pmusaraj (commented):
I have two questions here: 

1) How can we make sure that the topic timer that is being removed is the one added by the event reminder UI? I could be wrong, but it seems to me this can also remove non-calendar-related topic timers. 

2) Would be good to have tests here for a legitimately deleted timer as well as a non-event topic timer that doesn't get deleted when reminders of the event are edited.

PR #361, opened 2022-12-23
FIX: Make the “Interested” event button toggleable

jancernik:


assets/javascripts/discourse/widgets/discourse-post-event.js
```diff
@@ -91,10 +91,15 @@ export default createWidget("discourse-post-event", {
 
   changeWatchingInviteeStatus(status) {
     if (this.state.eventModel.watching_invitee) {
+      const currentStatus = this.state.eventModel.watching_invitee.status;
+      let newStatus = status;
+      if (status === currentStatus && status === "interested") {
+        newStatus = null;
+      }
```
ZogStriP, line 98: Why the need to create another variable? Does this not work? 🤔 

```suggestion
      if (status === currentStatus && status === "interested") {
        status = null;
      }
```

jancernik: It's not necessarily good practice to [reassign parameters](https://eslint.org/docs/latest/rules/no-param-reassign). I thought ESlint might complain, but I just tested, and it didn't, so I'm down for the change

ZogStriP: Right, objects are passed by reference in JS. In that case you could do something like this

```javascript
const newStatus = status === currentStatus && status === "interested" ? null : status;
```

jancernik: Yep, you can do that. Formated turns into 
```
      const newStatus =
        status === currentStatus && status === "interested" ? null : status;
```
which isn't the cleanest IMO

ZogStriP (approved):
Any chances you could also add a test?

jancernik:
It was my intention, but this falls in a weird place for acceptance tests for such a small thing. I'm going to merge this as is

PR #364, opened 2022-12-29
FEATURE: Add option to leave public events

jancernik:


app/controllers/discourse_post_event/invitees_controller.rb
```diff
@@ -45,7 +45,7 @@ def create
     def destroy
       event = Event.find_by(id: params[:post_id])
       invitee = event.invitees.find_by(id: params[:id])
-      guardian.ensure_can_act_on_discourse_post_event!(event)
+      guardian.ensure_can_act_on_invitee!(invitee)
```
pmusaraj, line 48: Can we add a test that ensures user can delete their own invitee record? I think the current tests in `invitees_controller_spec.rb` are checking that admin can delete and lurker cannot delete.

pmusaraj (changes requested):
This looks good, but a test is necessary IMO.

app/controllers/discourse_post_event/invitees_controller.rb
jancernik: Sure, I'll add one

pmusaraj (approved):
Thanks! LGTM

PR #369, opened 2023-01-02
DEV: Fix a spec tied to a specific year

CvX:
The spec checks specific 2022 holidays, so it started to fail as soon as we entered 2023. 😉 Freezing the time to 2022 fixes the problem.


ZogStriP (approved):
LGTM but can you explain what this is fixing exactly in the commit message?

PR #373, opened 2023-01-06
FIX: Make users_on_holiday return type consistent

s3lase:
Return an empty array instead of nil when the plugin store has not yet been populated with `users_on_holiday`.

At the moment, the `currently_away` report fails if the `UpdateHolidayUsernames` never gets the chance to update `users_on_holiday`.

### DIFF
--- a/spec/integration/curently_away_report_spec.rb
+++ b/spec/integration/curently_away_report_spec.rb
@@ -7,16 +7,25 @@ describe "currently_away report" do
   fab!(:user_2) { Fabricate(:user) }
   fab!(:group_1) { Fabricate(:group) }
 
-  before do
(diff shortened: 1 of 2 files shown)


spec/integration/curently_away_report_spec.rb
```diff
@@ -9,14 +9,27 @@
 
   before do
     group_1.add(user_1)
+  end
+
+  context "when users_on_holiday is not set" do
+    it "does not generated report with data" do
```
ZogStriP, line 15: Small typo

```suggestion
    it "does not generate report with data" do
```

ZogStriP (approved):

s3lase:
Thanks @ZogStriP

PR #375, opened 2023-01-10
FIX: Sometimes the event doesn't load after creating it

jancernik:
Screenshot from 2023-01-03 09-01-30

Currently, the `update_from_raw` function gets called by the `:post_process_cooked` event,. This sometimes causes the event not to load after creating it. I haven't been able to pinpoint the exact problem here, but my best bet is that it's some sort of timing-related issue, as it's fairly inconsistent to reproduce.

Since `update_from_raw` doesn't affect the cooked output, triggering it earlier solves the loading problem.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -254,7 +254,9 @@ after_initialize do
     SiteSetting.discourse_post_event_enabled && !object.nil? && !object.deleted_at.present?
   end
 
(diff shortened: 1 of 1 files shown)


pmusaraj (approved):
Thanks, let's give it a try!

PR #381, opened 2023-01-17
FEATURE: allow disabling automatic holidays globally

SamSaffron:
Automatic holidays can not be disabled by disabling
`calendar_automatic_holidays_enabled`

By default we will set automatic holidays

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -42,6 +42,7 @@ discourse_calendar:
   close_to_working_day_hours_extension:
     default: 2
     client: true
+  calendar_automatic_holidays_enabled: true
 
 discourse_post_event:
   discourse_post_event_enabled:
--- a/jobs/scheduled/create_holiday_events.rb
+++ b/jobs/scheduled/create_holiday_events.rb
@@ -5,8 +5,12 @@ module Jobs
     every 10.minutes
 
     def execute(args)
-      return unless SiteSetting.calendar_enabled
-      return unless topic_id = SiteSetting.holiday_calendar_topic_id.presence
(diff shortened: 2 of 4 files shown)


config/settings.yml
```diff
@@ -42,6 +42,7 @@ discourse_calendar:
   close_to_working_day_hours_extension:
     default: 2
     client: true
+  calendar_automatic_holidays_enabled: true
```
martin-brennan, line 45: Not sure this is right...should be:

```ruby
calendar_automatic_holidays_enabled:
  default: true
```

jobs/scheduled/create_holiday_events.rb
```diff
@@ -5,8 +5,12 @@ class ::DiscourseCalendar::CreateHolidayEvents < ::Jobs::Scheduled
     every 10.minutes
 
     def execute(args)
-      return unless SiteSetting.calendar_enabled
-      return unless topic_id = SiteSetting.holiday_calendar_topic_id.presence
+      return if !SiteSetting.calendar_enabled
```
martin-brennan, line 8: Yay only good unless is a dead unless :laughing:

config/settings.yml
martin-brennan: Maybe I am wrong and we allow that other syntax?

SamSaffron: yeah the other syntax is only required when you are using client: true... I prefer the more concise one where possible

martin-brennan: Right, carry on then!

martin-brennan (approved):

PR #385, opened 2023-01-24
FEATURE: Option to configure the holiday emoji

jancernik:
This allows setting a custom emoji for the holiday status and changes the default from :desert_island: to :date:

### DIFF
--- a/spec/integration/post_spec.rb
+++ b/spec/integration/post_spec.rb
@@ -612,6 +612,24 @@ describe Post do
           expect(status.ends_at).to eq_time(Time.utc(2018, 6, 6, 10, 20))
         end
       end
+
+      context "when custom emoji is blank" do
+        before { SiteSetting.holiday_status_emoji = "" }
+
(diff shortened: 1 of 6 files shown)


keegangeorge (approved):

spec/integration/post_spec.rb
```diff
@@ -592,6 +592,26 @@
         expect(status.description).to eq(custom_status[:description])
         expect(status.emoji).to eq(custom_status[:emoji])
       end
+
+      context "when custom emoji is set" do
+        custom_emoji = "palm_tree"
+
+        before { SiteSetting.holiday_status_emoji = custom_emoji }
+
+        it "sets holiday user status with custom emoji" do
```
pmusaraj, line 601: Can we add a test case here for when `holiday_status_emoji` is blank? I see that it will default to the date emoji, but a test case would be good just to prevent regressions in the future.

pmusaraj (approved):
Looks good, just a tiny comment on specs.

PR #392, opened 2023-03-13
FIX: Add backward compatibility for reminders

jancernik:
The migration did not change the raw topic content. Therefore, if you edit and save the event through the composer (not the modal), it applied back the old format. This change is to deal with those cases.

old syntax:   `reminders="<value>.<unit>"`
new syntax:  `reminders="<type>.<value>.<unit>"`


pmusaraj (approved):
Thanks!

PR #418, opened 2023-07-13
FIX: Missing event user notifications

jdmartinez1062:
**Context**
The calendar plugin allows the creation of events, and the notification of users of the existence of that event after it has been created

**Problem**
When trying to notify a user the notification never shows on the user's side. The methods involved in the notification seem to be missing information

**Solution**
Fix the methods involved in the notification flow


spec/models/discourse_post_event/event_spec.rb
CvX, line 1: Looks like this file needs a stree formatting pass 🙂

CvX (approved):

PR #429, opened 2023-08-17
FIX: Event sorting shows oldest events first

jdmartinez1062:
**Context**
The discourse calendar plugin allows the creation of events. These events can be sorted by the `created_at` date when the `sort categories by event start date enabled` `SiteSetting` is enabled.

**Problem**
When the `sort categories by event start date enabled` `SiteSetting` is enabled the events are sorted from oldest to recent making it very difficult for the user to navigate and find Topics of interest

**Solution**
Fix how events are sorted so events that are closer to happening appear first and events that already happened appear from oldest to most recent

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -111,8 +111,9 @@ after_initialize do
         if category && category.custom_fields &&
              category.custom_fields["sort_topics_by_event_start_date"]
           reorder_sql = <<~SQL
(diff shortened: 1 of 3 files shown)


ZogStriP (approved):

ZogStriP:
Although the specs error looks relevant.

PR #432, opened 2023-08-25
FEATURE: Display calendar events adjusted for timezones

jancernik:
This PR adds the option to enable a timezone adjustment for calendar events. This will make it so events render offset from the grid to reflect the appropriate start and end moments according to the viewer's timezone.

- Events will rerender when changing the timezone from the timezone picker
- The calendar needs to be full day (`fullDay="true"`)
- The hidden site setting is `enable_timezone_offset_for_calendar_events`

<div align="center"> <h3>Denver (GMT-6) / Tokyo (GMT+9)</h3>

calendar demo

</div>

By default, grouped events (holidays) get split by timezone. This can be configured. If the timezone difference between events is less than a set threshold, events get grouped anyway and will be displayed with the timezone closest to the average.

- The hidden site setting is `split_grouped_events_by_timezone_threshold`

<div align="center"> <h3>Threshold<br>3 / 1 / 0</h3>

threshold

</div>

### DIFF
--- a/assets/javascripts/discourse/initializers/discourse-calendar.js
+++ b/assets/javascripts/discourse/initializers/discourse-calendar.js
@@ -523,46 +523,111 @@ function initializeDiscourseCalendar(api) {
     calendar.addEvent(event);
   }
 
-  function _addGroupedEvent(calendar, post, detail) {
-    let htmlContent = "";
-    let usernames = [];
-    let localEventNames = [];
+  function _addGroupedEvent(calendar, post, detail, timezone) {
+    const data = siteSettings.enable_timezone_offset_for_calendar_events
+      ? _splitGroupEventByTimezone(detail, timezone)
+      : [detail];
 
-    Object.keys(detail.localEvents)
-      .sort()
-      .forEach((key) => {
-        const localEvent = detail.localEvents[key];
-        htmlContent += `<b>${key}</b>: ${localEvent.usernames
-          .sort()
-          .join(", ")}<br>`;
-        usernames = usernames.concat(localEvent.usernames);
-        localEventNames.push(key);
-      });
+    data.forEach((el) => {
+      let htmlContent = "";
+      let users = [];
+      let localEventNames = [];
 
-    const event = _buildEvent(detail);
-    event.classNames = ["grouped-event"];
+      Object.keys(el.localEvents)
+        .sort()
+        .forEach((key) => {
+          const localEvent = el.localEvents[key];
+          htmlContent += `<b>${key}</b>: ${localEvent.users
+            .map((u) => u.username)
+            .sort()
+            .join(", ")}<br>`;
+          users = users.concat(localEvent.users);
+          localEventNames.push(key);
+        });
 
-    if (usernames.length > 2) {
-      event.title = `(${usernames.length}) ${localEventNames[0]}`;
-    } else if (usernames.length === 1) {
-      event.title = usernames[0];
-    } else {
-      event.title = isMobileView
-        ? `(${usernames.length}) ${localEventNames[0]}`
-        : `(${usernames.length}) ` + usernames.join(", ");
-    }
+      const event = _buildEvent(el);
+      event.classNames = ["grouped-event"];
 
-    if (localEventNames.length > 1) {
-      event.extendedProps.htmlContent = htmlContent;
-    } else {
-      if (usernames.length > 1) {
+      if (users.length > 2) {
+        event.title = `(${users.length}) ${localEventNames[0]}`;
+      } else if (users.length === 1) {
+        event.title = users[0].username;
+      } else {
+        event.title = isMobileView
+          ? `(${users.length}) ${localEventNames[0]}`
+          : `(${users.length}) ` + users.map((u) => u.username).join(", ");
+      }
+
+      if (localEventNames.length > 1) {
         event.extendedProps.htmlContent = htmlContent;
       } else {
-        event.extendedProps.htmlContent = localEventNames[0];
+        if (users.length > 1) {
+          event.extendedProps.htmlContent = htmlContent;
+        } else {
+          event.extendedProps.htmlContent = localEventNames[0];
+        }
       }
-    }
 
-    calendar.addEvent(event);
+      calendar.addEvent(event);
+    });
+  }
+
+  function _splitGroupEventByTimezone(detail, timezone) {
+    const calendarUtcOffset = moment.tz(timezone).utcOffset();
+    let timezonesOffsets = [];
+    let splittedEvents = [];
+
+    Object.values(detail.localEvents).forEach((event) => {
+      event.users.forEach((user) => {
+        const userUtcOffset = moment.tz(user.timezone).utcOffset();
+        const timezoneOffset = (calendarUtcOffset - userUtcOffset) / 60;
+        user.timezoneOffset = timezoneOffset;
+        timezonesOffsets.push(timezoneOffset);
+      });
+    });
+
+    [...new Set(timezonesOffsets)].forEach((offset) => {
+      let filteredLocalEvents = {};
+      let eventTimezone;
+
+      Object.keys(detail.localEvents).forEach((key) => {
+        const filtered = detail.localEvents[key].users.filter(
+          (user) => user.timezoneOffset === offset
+        );
+        if (filtered.length > 0) {
+          filteredLocalEvents[key] = {
+            users: filtered,
+          };
+          eventTimezone ||= filtered[0].timezone;
+        }
+      });
+
+      let from = moment.tz(detail.from, eventTimezone);
+      let to = moment.tz(detail.to, eventTimezone);
+
+      if (offset > 0) {
+        if (to.isValid()) {
+          to.add(1, "day");
+        } else {
+          to = from.clone().add(1, "day");
+        }
+      } else if (offset < 0) {
+        if (!to.isValid()) {
+          to = from.clone();
+        }
+        from.subtract(1, "day");
+      }
+
+      splittedEvents.push({
+        timezoneOffset: offset,
+        localEvents: filteredLocalEvents,
+        eventDaysDuration: 1,
+        from: from.format("YYYY-MM-DD"),
+        to: to.format("YYYY-MM-DD"),
+      });
+    });
+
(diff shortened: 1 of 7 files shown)


pmusaraj:
@jancernik One issue I noticed with public holidays is that they can span two days. Ca you repro this? 

image

Her the grey colored event is Labor Day. User "testuser5" is in Los Angeles, user "pmusaraj" in Montreal (Canada). The holiday is only on the 4th, not sure why this is showing it as starting a day before. Curious if you can repro.

jancernik:
Hmm, thanks for letting me know. I'm working on this kind of issues right now.

I had a similar thing happen when testing in Chrome and Firefox. Due to differences in the way they handle the default Date object, with some timezones, all the events get rendered one day earlier in Firefox.

This is not related to this change. I found out it's always been this way.


<div align="center">



<b>Firefox / Chrome</b>

</div>

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -629,19 +741,23 @@ function initializeDiscourseCalendar(api) {
...
+        resetDynamicEvents();
         _insertAddToCalendarLinks(calendar);
       });
 
-      moment.tz.names().forEach((tz) => {
-        tzPicker.appendChild(new Option(tz, tz));
-      });
+      moment.tz
+        .names()
+        .filter((t) => !t.startsWith("Etc/GMT"))
```
pmusaraj, line 757: Curious, why do we need this?

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -329,15 +351,15 @@ function initializeDiscourseCalendar(api) {
     };
 
     let event = {
-      start: from.dateTime.toDate(),
+      start: from.dateTime.format("YYYY-MM-DD"),
```
pmusaraj, line 354: Is this going to drop the time of the event on the list view? I.e. this: 

image

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -663,6 +779,49 @@ function initializeDiscourseCalendar(api) {
...
+      info.view.type === "listNextYear"
+    ) {
+      return;
+    }
+
+    const timezoneOffset = info.event.extendedProps.timezoneOffset;
+    const segmentDuration = info.el.parentNode?.colSpan;
+
+    const basePctOffset = 100 / segmentDuration;
+    const basePxOffset = 5.5 - (segmentDuration - 1) * 0.78;
```
pmusaraj, line 794: This is a little bit too magic. Maybe add a comment explaining the reasoning for this?

plugin.rb
```diff
@@ -363,22 +363,23 @@ module ::DiscoursePostEvent
...
+          }
         end
       end
 
     grouped.each do |_, v|
       v[:name].sort!.uniq!
       v[:name] = v[:name].join(", ")
-      v[:usernames].sort!.uniq!
+      v[:users].sort! { |a, b| a[:username] <=> b[:username] }
+      v[:users].uniq! { |u| u[:username] }
```
pmusaraj, line 382: Is the backend test failure related to this?

pmusaraj (commented):
Nice progress, just a few minor comments, but overall I like the recent changes. 

I'd really like to merge and test this with real data shortly.  Do you think you can make some final fine-tuning and merge this today or Monday @jancernik ?

assets/javascripts/discourse/initializers/discourse-calendar.js
jancernik: I added that comment to the commit

These options (Etc/GMT-04, Etc/GMT+5, etc.) are being parsed incorrectly.
They are essentially being parsed as if the signs were flipped.
I tried forcefully feeding the timezone with the opposite sign,
which fixed the event rendering but not the calendar itself.
I'm opting to remove them from the list to avoid confusion.

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -514,50 +544,153 @@ function initializeDiscourseCalendar(api) {
...
+      });
+    });
+
+    [...new Set(timezonesOffsets)].forEach((offset) => {
+      let filteredLocalEvents = {};
+      let eventTimezones = [];
+
+      Object.keys(detail.localEvents).forEach((key) => {
+        const threshold =
+          siteSettings.split_grouped_events_by_timezone_threshold;
```
pmusaraj, line 617: I tried locally to set this to various numbers, but I no longer see the events being offset in the UI. Is there a regression? For example, `testuser5` is set to a timezone in Africa, but their events don't offset by ~5 hours, regardless of the threshold value (I tried 0, 3, 5). 

image

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -328,16 +350,20 @@ function initializeDiscourseCalendar(api) {
       return d.hours() !== 0 || d.minutes() !== 0 || d.seconds() !== 0;
     };
 
+    const hasTime =
+      hasTimeSpecified(to.dateTime) || hasTimeSpecified(from.dateTime);
```
pmusaraj, line 354: This line is causing a JS error locally when I have events without a `to` date. Calendar is empty and I see this: 

image

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -663,6 +826,63 @@ function initializeDiscourseCalendar(api) {
...
+    // original size while adjusting it to the respective timezone.
+
+    const timezoneOffset = info.event.extendedProps.timezoneOffset;
+    const segmentDuration = info.el.parentNode?.colSpan;
+
+    const basePctOffset = 100 / segmentDuration;
+    // Base margin required to shrink down the event by one day
+    const basePxOffset = 5.5 - segmentDuration;
+    // Default space between two consecutive events
+    // 5.5px = ( ( ( 2px margin + 3px padding ) * 2 ) + 1px border ) / 2
```
pmusaraj, line 850: This is nice, thanks!

assets/javascripts/discourse/initializers/discourse-calendar.js
jancernik: Does the other non-grouped events work? Do you have ` fullDay="true"`?
I don't think the new setting has much to do with the offset

pmusaraj: Here, I set up testuser5 as a user in Africa, so, different timezone. The markdown for the events for that user is this: 

```
[date=2023-09-28 timezone="Africa/Accra"]
```

THis shows as a day before in the calendar for a user with the timezone in Toronto. 

```
[date-range from=2023-09-30T10:00:00 to=2023-09-30T13:00:00 timezone="Africa/Accra"]
```

This shows as a full-day event on the 30th, no offset, and time range isn't respected AFAICT.

jancernik: I tried reproducing your issue, but it's working just fine for me.

This is a user with the timezone profile set to "Africa/Accra". Although that does not really matter much, what matters is the timezone attribute in the date. I posted both of your dates.

Viewing the calendar with a user from "America/Toronto" I see this:

image

My calendar markdown is `[calendar weekends=true tzPicker="true" fullDay="true"] [/calendar]`

And yes, the time range is not respected. That's why it That's why it needs `fullDay="true"`. Maybe in the future we can add a way to also adjust the event length according to time, but I left it out of the scope of this PR.

pmusaraj: Ahh, `fullDay=true` was the issue... with it, it's now fixed. Apologies, I missed that on your previous comment.

test/javascripts/acceptance/timezone-offset-test.js
```diff
@@ -0,0 +1,358 @@
...
+const topicResponse = {
+  post_stream: {
+    posts: [
+      {
+        id: 375,
+        name: null,
+        username: "jan",
+        avatar_template: "/letter_avatar_proxy/v4/letter/j/ce7236/{size}.png",
+        created_at: "2023-09-08T16:50:07.638Z",
+        cooked:
```
pmusaraj, line 14: Cool, good to see an acceptance test here! 

Maybe add the `raw: ` value as well? It's not going to affect the test, but makes this a bit more readable, i.e. this raw markdown will generate that markup and calendar.

test/javascripts/acceptance/timezone-offset-test.js
```diff
@@ -0,0 +1,358 @@
...
+    assert.notOk(eventElement.style.marginRight);
+  });
+
+  test("applies the correct offset for events that extend into the next day", async (assert) => {
+    await visit("/t/252");
+
+    const eventElement = getEventByText("Cordoba");
+
+    assert.ok(eventElement.style.marginLeft.includes("8.33")); // ( ( 1 - (-3) ) / 24 ) * 50%
+    assert.ok(eventElement.style.marginRight.includes("41.66")); // ( ( 24 - ( 1 - (-3) ) ) / 24 ) * 50%
```
pmusaraj, line 277: This is cool but I am a little worried it's going to be fragile. Different browsers may be off by a hair, and the test will fail. CSS changes will also likely make it fail (the 1 and -3 values come from the stylesheet, right?). 

I'd be ok with making this less precise. Either by checking that the margin matches the rounded value (i.e. == 8) or simply checking the presence of a margin? It's certainly not as accurate, but the test will be a bit more resilient.

pmusaraj (approved):
Looks great locally, I tested a bit further and it's pretty solid. 

I left one small comment on the acceptance test, feel free to address it in this PR or merge as is and keep an eye on test runs. 

Thanks @jancernik!

test/javascripts/acceptance/timezone-offset-test.js
jancernik: Yes, I agree, it can fail at some point, but checking only for the margin presense feels like we are only testing one conditional. Checking the round value is better. We need to parse the string to get the value, but it's not that bad.
I'll update the test before merging.

1 and -3 correspond to the timezones, Lisbon (GMT+1), Cordoba (GMT-3)

PR #437, opened 2023-09-12
FEATURE: Add user option to toggle timezone adjustment

jancernik:
- The button only appears when `enable_timezone_offset_for_calendar_events` is set to true.
- The option is not persistent; it resets each time the calendar is loaded.
- By default, the offset is disabled and is configurable via `default_timezone_offset_user_option`.
- Similar to changing the timezone, it takes some time to rerender the events, depending on the number of them.

demo

### DIFF
--- a/assets/javascripts/discourse/initializers/discourse-calendar.js
+++ b/assets/javascripts/discourse/initializers/discourse-calendar.js
@@ -833,10 +833,11 @@ function initializeDiscourseCalendar(api) {
     );
     const timezoneButton = document.createElement("button");
 
+    timezoneButton.title = "Toggle timezone offset";
     timezoneButton.classList.add(
       "timezone-offset-button",
       "btn",
-      "btn-small",
+      "btn-default",
       "btn-icon",
       "no-text"
     );
--- a/assets/javascripts/discourse/initializers/discourse-calendar.js
+++ b/assets/javascripts/discourse/initializers/discourse-calendar.js
@@ -13,6 +13,7 @@ import { createPopper } from "@popperjs/core";
 import { isNotFullDayEvent } from "../lib/guess-best-date-format";
 import { formatEventName } from "../helpers/format-event-name";
 import getURL from "discourse-common/lib/get-url";
+import { iconHTML } from "discourse-common/lib/icon-library";
 
 function loadFullCalendar() {
   return loadScript(
@@ -20,10 +21,7 @@ function loadFullCalendar() {
   );
 }
 
-let calendarPost;
-let calendar;
 let eventPopper;
-
 const EVENT_POPOVER_ID = "event-popover";
 
 function initializeDiscourseCalendar(api) {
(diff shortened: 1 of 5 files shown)


assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -295,11 +300,21 @@ function initializeDiscourseCalendar(api) {
...
-        left: "prev,next today",
+        left: "prev,next today timezoneOffset",
         center: "title",
         right: "month,basicWeek,listNextYear",
       },
       eventOrder: ["start", _orderByTz, "-duration", "allDay", "title"],
+      customButtons:
+        siteSettings.enable_timezone_offset_for_calendar_events && {
+          timezoneOffset: {
+            text: "Timezone Offset",
```
pmusaraj, line 311: Can we use the `globe` icon instead of a text label? It would take less space in the screen. You can try with either `iconNode("globe")` or `iconHTML("globel")` helpers.

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -30,6 +33,8 @@ function initializeDiscourseCalendar(api) {
     return;
   }
 
+  let enableTimezoneOffset = siteSettings.default_timezone_offset_user_option;
```
pmusaraj, line 36: Do we have too many site settings related to this feature now? Maybe worth a followup PR.

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -20,7 +20,10 @@ function loadFullCalendar() {
   );
 }
 
+let calendarPost;
+let calendar;
```
pmusaraj, line 24: I see a lot of removed `calendar` references. Do we still need to declare this here?

pmusaraj (approved):
Looks good, mostly cosmetic comments.

assets/javascripts/discourse/initializers/discourse-calendar.js
jancernik: The default icon support with the FullCalendar library is very limited. It doesn't allow custom icons or HTML. We would need to render it and then replace the innerHTML afterwards. But yes, I agree, an icon would be much better

assets/javascripts/discourse/initializers/discourse-calendar.js
jancernik: I wasn't thinking of adding this one, but since it takes a while to rerender the events when clicking the button, it makes it more reliable for testing. Maybe aother way could be to await for the the button state or something.
Yes, I'll probably have a followup PR revising all of this

assets/javascripts/discourse/initializers/discourse-calendar.js
jancernik: Yes, since I had to access the `calendar` object from a function defined within its creation, I had to make the variable global. That's why I removed many of the references when passing it as an attribute between functions, but we still need it

assets/javascripts/discourse/initializers/discourse-calendar.js
pmusaraj: Can we use our own button, outside the calendar, next to the timezone dropdown?

jancernik:
I changed the code to manually add a button in the header instead. I used `iconHTML("globe")` instead of text.
It's nice because it required less refactoring, so I reverted those changes back.
I updated the gif in the description.

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -810,6 +827,38 @@ function initializeDiscourseCalendar(api) {
...
   }
 
+  function _setupTimezoneOffsetButton(resetDynamicEvents) {
+    const timezoneWrapper = document.querySelector(
+      ".discourse-calendar-timezone-wrap"
+    );
+    const timezoneButton = document.createElement("button");
+
+    timezoneButton.classList.add("fc-timezoneOffset-button");
+    timezoneButton.classList.add("fc-button");
```
pmusaraj, line 837:
```suggestion
    timezoneButton.classList.add("fc-timezoneOffset-button", "btn", "btn-icon");
```

I'd recommend using the button classes from core ("btn" and "btn-default") here. It makes the button styling more consistent, and you shouldn't need to add an active state class (below) or specific SCSS styling, it will inherit from core. 

Not sure this works 100%. Make sure you also look at the styleguide, `http://localhost:4200/styleguide/atoms/buttons`, it has great examples of button types and button states.

pmusaraj (approved):
Looking good! Approving, but I do think the button CSS classes need a little bit of cleaning up to match core.

assets/javascripts/discourse/initializers/discourse-calendar.js
pmusaraj: Also, for accessibility, we should also add a `title` or an `ariaLabel` property to the button.

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -833,10 +833,11 @@ function initializeDiscourseCalendar(api) {
     );
     const timezoneButton = document.createElement("button");
 
+    timezoneButton.title = "Toggle timezone offset";
```
pmusaraj, line 836: This needs to be localizable, something like: 

```javascript
I18n.t("discourse_calendar.toggle_timezone_offset_title");
```

and then `toggle_timezone_offset_title` should be added to client.en.yml.

pmusaraj (approved):
Sweet, LGTM!

PR #440, opened 2023-09-28
FEATURE: allow every 4 weeks as an interval

SamSaffron:
This allows us to sync up 2 week and 4 week events, monthly goes out of
sync

Also fixes... test suite failing in Australia

### DIFF
--- a/lib/discourse_post_event/rrule_generator.rb
+++ b/lib/discourse_post_event/rrule_generator.rb
@@ -7,20 +7,7 @@ class RRuleGenerator
     tzid ||= "UTC"
 
     rrule = generate_hash(base_rrule)
-    p rrule
     rrule = set_mandatory_options(rrule, starts_at)
-
-    p rrule
-    p starts_at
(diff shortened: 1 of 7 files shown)


lib/discourse_post_event/rrule_generator.rb
```diff
@@ -3,10 +3,25 @@
...
 class RRuleGenerator
-  def self.generate(base_rrule, starts_at, tzid: "UTC")
+  def self.generate(base_rrule, starts_at, tzid: nil)
+    tzid ||= "UTC"
+
     rrule = generate_hash(base_rrule)
+    p rrule
     rrule = set_mandatory_options(rrule, starts_at)
 
+    p rrule
```
jjaffeux, line 13: dis you mean to keep all this debugging?

SamSaffron: lol oops

ZogStriP (approved):

PR #441, opened 2023-09-28
FEATURE: Use event category's color for calendar event

nattsw:
We've hardcoded the events to `var(--tertiary)`. Many users have requested that events be the colour of their categories.

I attempted to add an acceptance test to assert that the colours match the category colours defined in the fixtures, but 
```
query(".fc-event-container").style.backgroundColor
```

returns `'rgb(37, 170, 226)'` value which is slightly annoying to assert against...

Category calendar now:
Screenshot 2023-09-28 at 6 41 36 PM

Upcoming calendar now:
Screenshot 2023-09-28 at 6 53 56 PM

### DIFF
--- a/app/serializers/discourse_post_event/event_serializer.rb
+++ b/app/serializers/discourse_post_event/event_serializer.rb
@@ -27,6 +27,7 @@ module DiscoursePostEvent
     attributes :reminders
     attributes :recurrence
     attributes :minimal
+    attributes :category_color
 
     def can_act_on_discourse_post_event
       scope.can_act_on_discourse_post_event?(object)
@@ -127,6 +128,10 @@ module DiscoursePostEvent
     def should_display_invitees
       (object.public? && object.invitees.count > 0) ||
         (object.private? && object.raw_invitees.count > 0)
+    end
+
+    def category_color
+      object.post.topic.category.color
     end
   end
 end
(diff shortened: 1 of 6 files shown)


app/serializers/discourse_post_event/event_serializer.rb
```diff
@@ -128,5 +129,9 @@ def should_display_invitees
       (object.public? && object.invitees.count > 0) ||
         (object.private? && object.raw_invitees.count > 0)
     end
+
+    def category_color
+      object.post.topic.category.color
```
tgxworld, line 134: Just confirming here that `post`, `topic` and `category` has been eager loaded? Otherwise, this is going to result in the N+1 queries problem. 

Even if we did eager load stuff, I'm wondering how much potential overhead we might be introducing here as there may be alot of events being serialized.

tgxworld: Also, do we need to check if the user can see the category before we "reveal" its color?

tgxworld:
> returns 'rgb(37, 170, 226)' value which is slightly annoying to assert against...

I think this is fine given that for a given `hex` value we can only get a certain `rgb` value.

app/serializers/discourse_post_event/event_serializer.rb
nattsw: Resolved this by not loading the category, using `topic.category_id` and the category store

renato (approved):

PR #445, opened 2023-10-03
DEV: Update QUnit acceptance test to not rely on legacy nav menu

tgxworld:
Why this change?

The legacy navigation menu has been deprecated for quite awhile and will
soon be removed.

What does this change do?

Update the tests to test against the `sidebar` navigation menu instead
which is the default for a new Discourse install.


lis2 (approved):

tgxworld:
Thank you @lis2

PR #448, opened 2023-10-07
FIX: Calendar events not rendering in Firefox

jancernik:
`new Date("2023-10-12 07:00:00-03:00")` -> `Date Thu Oct 12 2023 07:00:00 GMT-0300`
`new Date("2023-10-12 07:00:00 -03:00")` -> `Invalid Date`


CvX (approved):

CvX:
(would be awesome if we had a test covering displaying events like that 😃)

PR #460, opened 2023-10-20
DEV: Switch to new `addComposerToolbarPopupMenuOption` plugin API

tgxworld:
Why this change?

`api.addToolbarPopupMenuOptionsCallback` has been deprecated in https://github.com/discourse/discourse/commit/913fd3a7b392b492f6344102577960a6eada00ce


test/javascripts/acceptance/post-event-builder-test.js
```diff
@@ -21,7 +21,7 @@ acceptance("Post event - composer", function (needs) {
     await categoryChooser.expand();
     await categoryChooser.selectRowByValue(2);
     await click(".toolbar-popup-menu-options .dropdown-select-box-header");
-    await click(".toolbar-popup-menu-options *[data-value='insertEvent']");
+    await click(".toolbar-popup-menu-options *[data-name='Create event']");
```
ZogStriP, line 24: Would it be worth using `I18n.t` here instead of hard coding the text?

PR #461, opened 2023-10-24
FIX: fallback category color if there's no category

renato:
Fallback to a default color if there's no category_id.

A user may manually copy the event from a topic or convert one to a PM, leaving the event without a category.


nattsw (approved):
LGTM 👍 

Screenshot 2023-10-24 at 2 38 08 PM

PR #463, opened 2023-10-24
FEATURE: upcoming events list component

renato:
Adds an Upcoming events list component, which uses the current category route to display a list of upcoming events.

image

When requesting the category events fail:
image

It is intended to be used with the [Right Sidebar Blocks](https://meta.discourse.org/t/right-sidebar-blocks/231067) theme component.

When adding the component to the Right Sidebar TC, the parameters `monthFormat` `dateFormat` and `timeFormat` can be overridden with a [different one](https://momentjs.com/docs/#/displaying/format/).


assets/javascripts/discourse/services/post-event-api.js
```diff
@@ -0,0 +1,30 @@
+import Service from "@ember/service";
+import { ajax } from "discourse/lib/ajax";
+
+/**
+ * Discourse Post Event API Service
+ *
+ * @implements {@ember/service}
+ */
+export default class PostEventApi extends Service {
```
pmusaraj, line 9: I'm not sure this should be a service... it feels like it would be noisy if ajax calls were are following the same pattern. Not sure services are an appropriate vehicle for this, I think it's fine in this case to have the ajax call in the component.

test/javascripts/integration/components/upcoming-events-list-test.gjs
```diff
@@ -0,0 +1,282 @@
...
+  setupRenderingTest(hooks);
+
+  let clock;
+
+  hooks.beforeEach(function () {
+    this.owner.unregister("service:router");
+    this.owner.register("service:router", RouterStub);
+
+    // this.siteSettings.discourse_calendar_enabled = true;
+    // this.siteSettings.discourse_post_event_enabled = true;
```
pmusaraj, line 37: You probably don't need these two commented-out lines.

assets/javascripts/discourse/components/upcoming-events-list.gjs
```diff
@@ -0,0 +1,153 @@
...
+
+  get hasEmptyResponse() {
+    return !this.isLoading && Object.keys(this.eventsByMonth).length === 0;
+  }
+
+  @action
+  async updateEventsByMonth() {
+    this.isLoading = true;
+
+    const { events } = await this.postEventApi.categoryEvents(this.categoryId);
```
pmusaraj, line 60: This might need a `try / catch`. What would happen if this request fails? We don't necessarily want to show a modal, but the component might end up in loading state forever?

pmusaraj (commented):
Looks generally great, just a couple of small comments.

test/javascripts/integration/components/upcoming-events-list-test.gjs
renato: Yup, leftover, will remove 👍

assets/javascripts/discourse/components/upcoming-events-list.gjs
renato: Yeah, I missed that. Maybe something like "Failed to retrieve events. [Try again]" should be enough?

renato:
Thanks for the review @pmusaraj!

assets/javascripts/discourse/components/upcoming-events-list.gjs
pmusaraj: Sure, yeah, that makes sense.

assets/javascripts/discourse/services/post-event-api.js
pmusaraj: That's a good question, I'm not sure we currently have an official pattern. Some core API calls are in the model, for example the `user` model has a list. 

Might be worth a longer internal discussion.

renato:
Hey @pmusaraj, it's ready for another round when you're available. Thank you!

pmusaraj (approved):
Looks great, thanks, LGTM!

PR #476, opened 2023-11-09
DEV: Update Event Invitees modal to new Component API

janzenisaac:
- Upgrade modal to new component based API
- Remove redundant event status in RSVPed modal

# Before
Screenshot 2023-11-09 at 12 28 03 PM

# After
Screenshot 2023-11-09 at 1 12 58 PM


assets/javascripts/discourse/components/modal/discourse-post-event-invitees.js
```diff
@@ -0,0 +1,65 @@
+import Component from "@glimmer/component";
+import { tracked } from "@glimmer/tracking";
+import { action } from "@ember/object";
+import { inject as service } from "@ember/service";
+import { debounce } from "discourse-common/utils/decorators";
+
+export default class DiscoursePostEventInvitees extends Component {
```
CvX, line 7: btw, we don't need the `Discourse` prefix in the name? could be a good opportunity to drop that 😃

assets/javascripts/discourse/components/modal/discourse-post-event-invitees.js
```diff
@@ -0,0 +1,65 @@
...
+  constructor() {
+    super(...arguments);
+    this._fetchInvitees();
+  }
+
+  get title() {
+    return `discourse_post_event.invitees_modal.${
+      this.args.model.params?.title || "title_invited"
+    }`;
+  }
```
CvX, line 24: minor change but I'd use `I18n.t()` here, so that the getter would actually return the title (so the name would match the impl) and not title's i18n path

assets/javascripts/discourse/widgets/discourse-post-event.js
```diff
@@ -32,20 +32,13 @@ export default createWidget("discourse-post-event", {
...
-        model: eventModel,
-        title: `discourse_post_event.invitees_modal.${title}`,
-        modalClass: [`${dasherize(name).toLowerCase()}-modal`, extraClass].join(
-          " "
-        ),
+    this.store
+      .find("discourse-post-event-event", params.postId)
+      .then((eventModel) => {
+        this.modal.show(DiscoursePostEventInvitees, {
+          model: { event: eventModel, params },
```
CvX, line 39: I'd pass `title` and `extraClass` rather than whole `params` (even just to avoid super deep `this.args.model.params.[prop]` in the component 😅)

assets/javascripts/discourse/components/modal/post-event-invitees.js
```diff
@@ -0,0 +1,63 @@
+import Component from "@glimmer/component";
+import { tracked } from "@glimmer/tracking";
+import { action } from "@ember/object";
+import { inject as service } from "@ember/service";
+import { debounce } from "discourse-common/utils/decorators";
+import I18n from "I18n";
```
CvX, line 6: One more change that happened when you were away 😄 

```suggestion
import I18n from "discourse-i18n";
```

CvX (approved):

PR #479, opened 2023-11-13
DEV: Update bulk-invite modal to Component API

janzenisaac:
- Update bulk-invite modal to component based API

- Fix input styles
### Before 
Screenshot 2023-11-13 at 12 07 32 PM

### After
Screenshot 2023-11-13 at 11 49 45 AM


assets/javascripts/discourse/components/modal/post-event-bulk-invite.js
```diff
@@ -0,0 +1,113 @@
...
+    this.bulkInvites.pushObject(
+      EmberObject.create({ identifier: null, attendance })
+    );
+  }
+
+  @action
+  uploadDone() {
+    this.dialog.alert(I18n.t("discourse_post_event.bulk_invite_modal.success"));
+    this.args.closeModal();
+  }
```
janzenisaac, line 101: We have a race condition between closing the modal (and destroying the component) and then Uppy trying to call `uploadDone`. 

What if we did something like

```suggestion
  @action
  async uploadDone() {
    await this.dialog.alert(I18n.t("discourse_post_event.bulk_invite_modal.success"));
    this.args.closeModal();
  }
```

janzenisaac: This looks to be an uppy-upload issue in core. I will open a fix in a separate PR.

assets/javascripts/discourse/components/modal/post-event-bulk-invite.js
```diff
@@ -0,0 +1,113 @@
...
+
+  @action
+  removeBulkInvite(bulkInvite) {
+    this.bulkInvites.removeObject(bulkInvite);
+
+    if (!this.bulkInvites.length) {
+      this.bulkInvites = [
+        EmberObject.create({ identifier: null, attendance: "unknown" }),
+      ];
```
davidtaylorhq, line 84: Assigning like this means that it's no longer a TrackedArray 👀

davidtaylorhq (approved):

PR #480, opened 2023-11-13
DEV: Upgrade post-event-builder modal to Component API

janzenisaac:


assets/javascripts/discourse/components/modal/post-event-builder.js
```diff
@@ -0,0 +1,202 @@
...
+
+  @action
+  onChangeStatus(newStatus) {
+    this.args.model.updateEventRawInvitees([]);
+    if (newStatus === "private") {
+      this.args.model.updateEventRawInvitees(
+        this.args.model.event.raw_invitees.filter((x) => x !== "trust_level_0")
+      );
+    }
+    this.args.model.updateEventStatus(newStatus);
```
janzenisaac, line 95: This is basically copy pasted, and it doesn't make sense imo. I think it could be something like

```suggestion
    this.args.model.updateEventRawInvitees([]);
    this.args.model.updateEventStatus(newStatus);
```

davidtaylorhq (approved):

PR #483, opened 2023-11-17
FEATURE: Show all upcoming events for the next year

lis2:
Display upcoming recurrent events for a year.
In addition, download calendar is creating recurrent entries.

https://github.com/discourse/discourse-calendar/assets/72780/bf2261e6-cab3-4b5f-81c2-c4c102e66b65

Related PR https://github.com/discourse/discourse/pull/24404


app/serializers/discourse_post_event/event_serializer.rb
```diff
@@ -129,8 +131,50 @@ def should_display_invitees
...
     def category_id
       object.post.topic.category_id
     end
+
+    def include_next_dates?
+      object.recurrence
+    end
+
+    def next_dates
+      how_many_recurring_events =
```
martin-brennan, line 144: I think it will be beneficial to extract this to the `RRuleGenerator.generate` method via an additional parameter, `max_years: N`. Then you could have an internal method there, `def how_many_recurring_events(max_years)` that does the calculation you already have here times `max_years`.

This way you remove complexity from the serializer and end up with:

```ruby
RRuleGenerator.generate(
  recurrence_rule,
  object.starts_at.in_time_zone(object.timezone),
  tzid: object.timezone,
  max_years: 1
).map { |date| { starts_at: date, ends_at: date + difference.seconds } }
```

app/serializers/discourse_post_event/event_serializer.rb
```diff
@@ -129,8 +131,50 @@ def should_display_invitees
...
+
     def category_id
       object.post.topic.category_id
     end
+
+    def include_next_dates?
+      object.recurrence
+    end
+
+    def next_dates
```
martin-brennan, line 143: I think a _slightly_ better name for this is `upcoming_dates`.

app/serializers/discourse_post_event/event_serializer.rb
```diff
@@ -129,8 +131,50 @@ def should_display_invitees
       (object.public? && object.invitees.count > 0) ||
         (object.private? && object.raw_invitees.count > 0)
     end
+
     def category_id
       object.post.topic.category_id
     end
+
+    def include_next_dates?
+      object.recurrence
```
martin-brennan, line 140: I think you could add a nice `recurring?` method on `DiscoursePostEvent::Event` model that just does `object.recurrence.present?`, then you can call it like this:

```suggestion
      object.recurring?
```

app/serializers/discourse_post_event/event_serializer.rb
```diff
@@ -129,8 +131,50 @@ def should_display_invitees
...
+          recurrence_rule,
+          object.starts_at.in_time_zone(object.timezone),
+          tzid: object.timezone,
+        )
+        .first(how_many_recurring_events)
+        .map { |date| { starts_at: date, ends_at: date + difference.seconds } }
+    end
+
+    def include_recurrence_rule?
+      object.recurrence
```
martin-brennan, line 173: Same comment about `object.recurring?`

app/models/discourse_post_event/event.rb
```diff
@@ -398,13 +392,27 @@ def calculate_next_date
         byday = localized_start.strftime("%A").upcase[0, 2]
         recurrence = "FREQ=WEEKLY;BYDAY=#{byday}"
       end
+      recurrence
+    end
+
+    def calculate_next_date(start_date: nil)
+      if !start_date &&
+           (!original_ends_at || self.recurrence.blank? || original_starts_at > Time.current)
```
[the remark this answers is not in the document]
lis2: Good point! Renato touched it as well, so I will wait for his PR to be merged and then will take another look 

https://github.com/discourse/discourse-calendar/pull/485/files#diff-4ba5e47577238d6574a99940ee585162af91255e29f2490cd262dd4758edf73fR364

martin-brennan (approved):

martin-brennan:
Nice, thanks for making the fixes  :)

PR #486, opened 2023-11-21
FIX: regional public holidays day early

lis2:
When a public holiday is created, it is created in a specific user time zone.

https://github.com/discourse/discourse-calendar/blob/main/jobs/scheduled/create_holiday_events.rb#L71

When calendar is set as fullDay, we would like to display Christmas always 25th of December.

To achieve that, we need to have information about timezone, parse date in that timezone, and leave only year, month and day.

I added a test case to cover one bug scenario. User from Los Angeles have seen that user from Italy was celebrating Christmas Day on 24th of December instead of 25th.

### DIFF
--- a/assets/javascripts/discourse/initializers/discourse-calendar.js
+++ b/assets/javascripts/discourse/initializers/discourse-calendar.js
@@ -721,6 +721,11 @@ function initializeDiscourseCalendar(api) {
     (post.calendar_details || []).forEach((detail) => {
       switch (detail.type) {
         case "grouped":
+          if (fullDay && detail.timezone) {
+            detail.from = moment
+              .tz(detail.from, detail.timezone)
+              .format("YYYY-MM-DD");
+          }
           groupedEvents.push(detail);
           break;
         case "standalone":
--- a/jobs/scheduled/create_holiday_events.rb
+++ b/jobs/scheduled/create_holiday_events.rb
@@ -75,14 +75,18 @@ module Jobs
                   holiday[:date]
                 end
 
-              CalendarEvent.find_or_create_by(
-                topic_id: topic_id,
-                user_id: user_id,
-                username: usernames[user_id],
-                description: holiday[:name],
-                start_date: date,
-                region: region,
-              )
+              event =
+                CalendarEvent.find_or_initialize_by(
+                  topic_id: topic_id,
+                  user_id: user_id,
+                  username: usernames[user_id],
+                  description: holiday[:name],
+                  start_date: date,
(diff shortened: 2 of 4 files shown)


test/javascripts/acceptance/topic-calendar-holidays-test.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  test("renders calendar holidays with fullDay='true'", async (assert) => {
+    await visit("/t/-/252");
+
+    assert
+      .dom(".fc-week:nth-child(5) .fc-content-skeleton tbody td:first-child")
+      .hasClass(
+        "fc-event-container",
+        "Italian Christmas Day is displayed on Monday 2023-12-25"
```
nattsw, line 32: Interesting. I was wondering how "Italian" was decided, then saw the fixture below. Would it be more obvious if you make the fixture additions in line 21 of this file instead? (i.e. import then add to the array within this test file)

jobs/scheduled/create_holiday_events.rb
```diff
@@ -75,14 +75,18 @@ def execute(args)
...
+                CalendarEvent.find_or_create_by(
+                  topic_id: topic_id,
+                  user_id: user_id,
+                  username: usernames[user_id],
+                  description: holiday[:name],
+                  start_date: date,
+                  region: region,
+                )
+
+              # TODO: Merge update! to find_or_create_by after 06-2024.
```
nattsw, line 88: What is the context of this todo? Would you be creating a topic + timer for it?

assets/javascripts/discourse/initializers/discourse-calendar.js
```diff
@@ -721,6 +721,10 @@ function initializeDiscourseCalendar(api) {
     (post.calendar_details || []).forEach((detail) => {
       switch (detail.type) {
         case "grouped":
+          if (fullDay && detail.timezone) {
+            let from = moment.tz(detail.from, detail.timezone);
+            detail.from = from.format("YYYY-MM-DD");
```
nattsw, line 726:
```suggestion
            detail.from = moment
              .tz(detail.from, detail.timezone)
              .format("YYYY-MM-DD");
```

Is this legal? Also curious why the `let`

nattsw (approved):
LGTM with comments.

jobs/scheduled/create_holiday_events.rb
lis2: I had brain fog :)

PR #488, opened 2023-11-23
FIX: prevent holiday duplicates when username is changed

lis2:
When username was changed, the scheduled job was creating duplicated holidays.

In addition to fix, migration was added to remove duplicated and incorrect records.

### DIFF
--- a/db/migrate/20231123233308_delete_duplicated_holidays.rb
+++ b/db/migrate/20231123233308_delete_duplicated_holidays.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class DeleteDuplicatedHolidays < ActiveRecord::Migration[7.0]
+  def up
+    execute <<~SQL
+      DELETE
+        FROM calendar_events ce
+      WHERE
+        ce.id IN (SELECT ce2.id FROM calendar_events ce2
+                  INNER JOIN users ON users.id = ce2.user_id
+                  WHERE ce2.post_id IS NULL
+                    AND ce2.username != users.username)
+    SQL
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/spec/jobs/scheduled/create_holiday_events_spec.rb
+++ b/spec/jobs/scheduled/create_holiday_events_spec.rb
@@ -111,6 +111,17 @@ describe DiscourseCalendar::CreateHolidayEvents do
     expect(CalendarEvent.pluck(:region, :description, :start_date, :username)).to eq([])
   end
 
+  it "does not create duplicates when username is changed" do
+    frenchy
+    DiscourseCalendar::CreateHolidayEvents.new.execute(nil)
+    frenchy.update!(username: "new_username")
+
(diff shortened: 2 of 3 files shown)


spec/jobs/scheduled/create_holiday_events_spec.rb
```diff
@@ -111,6 +111,17 @@
     expect(CalendarEvent.pluck(:region, :description, :start_date, :username)).to eq([])
   end
 
+  it "does not create duplicates when username is changed" do
+    frenchy
+    DiscourseCalendar::CreateHolidayEvents.new.execute(nil)
+    frenchy.update!(username: "new_username")
```
martin-brennan, line 117: I think we just need another assertion here:

```ruby
created_event = CalendarEvent.last
expect(created_event.username).to eq(frenchy.username)
```

spec/jobs/scheduled/create_holiday_events_spec.rb
```diff
@@ -111,6 +111,17 @@
...
+
+    expect { DiscourseCalendar::CreateHolidayEvents.new.execute(nil) }.not_to change {
+      CalendarEvent.count
+    }
+    expect(CalendarEvent.last.username).to eq("new_username")
```
martin-brennan, line 122: Then this change:

```suggestion
    expect(created_event.reload.username).to eq("new_username")
```

db/migrate/20231123233308_delete_duplicated_holidays.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class DeleteDuplicatedHolidays < ActiveRecord::Migration[7.0]
+  def up
+    execute <<~SQL
+      DELETE
+        FROM calendar_events ce
+      WHERE
+        ce.id IN (SELECT ce2.id FROM calendar_events ce2
+                  INNER JOIN users ON users.id = ce2.user_id
```
martin-brennan, line 10: Is it possible for a different user to create the event for someone else?

db/migrate/20231123233308_delete_duplicated_holidays.rb
```diff
@@ -0,0 +1,19 @@
...
+                  WHERE ce2.post_id IS NULL
+                    AND ce2.username != users.username)
```
martin-brennan, line 12: Do we have to worry about case insensitivity here?

db/migrate/20231123233308_delete_duplicated_holidays.rb
lis2: No, Calendar events without post id are holidays only generated by scheduled job

db/migrate/20231123233308_delete_duplicated_holidays.rb
lis2: I think it is ok. Right now, if someone change username from `kris` to `Kris`, we would have duplicated holiday. We still want to delete `kris` and display properly `Kris`

martin-brennan (approved):

PR #489, opened 2023-11-24
FIX: remove duplicated holidays because of timezone change

lis2:
When a user updates timezone, it creates duplicated holidays.

Beside the fix, migration was added to remove duplicated records. Timezone change can only update date for 24 hours, so we are searching for records with start date around that range.


martin-brennan (approved):
LGTM!

PR #501, opened 2023-12-11
Show topic title suffix even when missing an ending

mbauman:
Addresses https://meta.discourse.org/t/post-event-date-not-always-displayed-on-topic-title/258816/ by using the start datetime. This won't ever show the event as being "in progress", but seems like a reasonable limitation.


ZogStriP:
Thanks for the PR @mbauman - I've made another one with tests -> https://github.com/discourse/discourse-calendar/pull/562

PR #507, opened 2023-12-15
FIX: events endpoint extension

renato:
~~Uses the extension parameter of [`appendQueryParams`](https://github.com/discourse/discourse/blob/48193767bf753bedf9483b4a0c3b2f9760745df3/app/assets/javascripts/discourse/app/adapters/rest.js#L47) instead of appending it manually.~~

Removes all `.json` extension occurrences from all `/discourse-post-event/events` calls.


janzenisaac:
Test failures look legit :)

janzenisaac (approved):

PR #511, opened 2023-12-20
FIX: site serializer should always return users_on_holiday

nlalonde:
Needs discussion. /t/116586/11


nlalonde:
We can't do this for privacy reasons.

PR #512, opened 2023-12-21
Fix holidays not showing in group timezones

nlalonde:


app/serializers/user_timezone_serializer.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
 class UserTimezoneSerializer < BasicUserSerializer
-  attributes :timezone
+  attributes :timezone, :on_holiday
+
+  def on_holiday
```
AndrewPrigorshnev, line 6: The name of this class is `UserTimezoneSerializer`, so `on_holiday` field doesn't fit in here, it has nothing to do with timezones. We either need to rename the class or we need to find another place for it.

AndrewPrigorshnev (approved):
This looks good.

But longterm we may want to consider passing to the client user status instead. That will be more correct solution because the calendar plugin relies on user status in Core since ccb4993e.

PR #529, opened 2024-01-18
DEV: fix another time dependent flaky test

renato:
Use explicit `starts_at` times for fabricated events to avoid flakiness


renato:
Looking into the test failure...

PR #531, opened 2024-01-22
DEV: Update public holiday for Singapore 2024

tgxworld:
https://www.mom.gov.sg/newsroom/press-releases/2023/0524-public-holidays-for-2024


vendor/holidays/test/defs/test_defs_sg.rb
```diff
@@ -7,19 +7,5 @@
 class SgDefinitionTests < Test::Unit::TestCase  # :nodoc:
 
   def test_sg
```
nattsw, line 9: I would prefer to remove the test file. But won't block this PR for it.

PR #533, opened 2024-01-24
DEV: Exclude system users when serializing group timezones

Drenmi:
### What is this change?

The `post_serializer_spec` was calling `Group.refresh_automatic_groups!(:admins)`, relying on that call booting system users from the group.

This change makes it so the implementation excludes any system users instead.


martin-brennan (approved):
NOICE

PR #549, opened 2024-03-14
FEATURE: use new options from downloadCalendar

jjaffeux:
"Add to calendar" will now prefill location with the url if present, and details with the associated post url, when using google calendar.

This has no tests as the feature is currently not tested at all in discourse-calendar and `downloadCalendar` is well tested in core.

Note: Im not adding a discourse compat entry as the current entry is already on the current discourse version.


assets/javascripts/discourse/widgets/discourse-post-event.js
```diff
@@ -178,7 +179,11 @@ export default createWidget("discourse-post-event", {
           endsAt: event.ends_at,
         },
       ],
-      event.recurrence_rule
+      {
+        recurrenceRule: event.recurrence_rule,
+        location: event.url,
+        details: getAbsoluteURL(event.post.url),
```
davidtaylorhq, line 185: Does this include the hostname? Might need to prepend `document.location.origin`? 🤔

PR #559, opened 2024-04-11
Update map_event_tag_colors_json_schema.rb

tobiaseigen:
Clarified that only hex codes are supported as  color


CvX (approved):

tobiaseigen:
no longer relevant as Kris has changed the color chooser to use the color wheel.

PR #562, opened 2024-05-06
FIX: support endless events in topic list

ZogStriP:
Uses the start date when an event doesn't have an end date so that we always show the date in the topic title in topic lists.

Cf. https://meta.discourse.org/t/post-event-date-not-always-displayed-on-topic-title/258816

Inspired by https://github.com/discourse/discourse-calendar/pull/501

### DIFF
--- a/assets/javascripts/discourse/components/event-date.gjs
+++ b/assets/javascripts/discourse/components/event-date.gjs
@@ -40,8 +40,7 @@ export default class EventDate extends Component {
   get shouldRender() {
     return (
       this.siteSettings.discourse_post_event_enabled &&
-      this.args.topic.event_starts_at &&
-      this.args.topic.event_ends_at
+      this.args.topic.event_starts_at
     );
   }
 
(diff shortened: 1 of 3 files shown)


assets/javascripts/discourse/components/event-date.gjs
```diff
@@ -50,11 +49,13 @@ export default class EventDate extends Component {
   }
 
   get eventEndedAt() {
```
Drenmi, line 51: I wonder if we should give this a better name now?

Drenmi (approved):
LGTM! 🚀

assets/javascripts/discourse/components/event-date.gjs
ZogStriP: Naming’s hard. I’m not sure what would be better 🤔 Technically, an event with just a start date ends up have his end date == start date. So it’s not _wrong_ to call it “end date”.

PR #564, opened 2024-05-13
FEATURE: closed events

ZogStriP:
Adds support for "closed" events instead of just relying on the start & end dates.

An event will be "expired" when it happened in the past.

An event might now also be "closed", if for whatever reasons, the author decides to "cancel" it.

Context - https://meta.discourse.org/t/when-closing-event-it-moves-it-to-todays-date-time/307292

### DIFF
--- a/app/serializers/discourse_post_event/event_serializer.rb
+++ b/app/serializers/discourse_post_event/event_serializer.rb
@@ -2,33 +2,34 @@
 
 module DiscoursePostEvent
   class EventSerializer < ApplicationSerializer
-    attributes :id
-    attributes :creator
-    attributes :sample_invitees
-    attributes :watching_invitee
-    attributes :starts_at
-    attributes :ends_at
-    attributes :timezone
-    attributes :stats
-    attributes :status
-    attributes :raw_invitees
-    attributes :post
-    attributes :name
     attributes :can_act_on_discourse_post_event
     attributes :can_update_attendance
+    attributes :category_id
+    attributes :creator
+    attributes :custom_fields
+    attributes :ends_at
+    attributes :id
+    attributes :is_closed
     attributes :is_expired
     attributes :is_ongoing
-    attributes :should_display_invitees
-    attributes :url
-    attributes :custom_fields
-    attributes :is_public
     attributes :is_private
+    attributes :is_public
     attributes :is_standalone
-    attributes :reminders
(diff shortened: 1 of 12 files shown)


app/serializers/discourse_post_event/event_serializer.rb
```diff
@@ -2,33 +2,34 @@
 
 module DiscoursePostEvent
   class EventSerializer < ApplicationSerializer
-    attributes :id
```
ZogStriP, line 5: I just sorted alphabetically all the attributes.

Only added "`is_closed`"

assets/javascripts/discourse/widgets/discourse-post-event-dates.js
```diff
@@ -15,27 +15,28 @@ export default createWidget("discourse-post-event-dates", {
...
     const content = [
       iconNode("clock"),
-      h("span.date", new RawHtml({ html: `<span>${attrs.localDates}</span>` })),
+      h("span.date", new RawHtml({ html: `<span>${localDates}</span>` })),
     ];
 
     if (
-      attrs.eventModel.is_expired &&
-      attrs.eventModel.status !== "standalone"
+      eventModel.is_expired &&
```
ZogStriP, line 25: Ensures we only show participants when the event was not "standalone" nor closed.

spec/system/post_event_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    composer.submit
+
+    expect(page).to have_content(title)
+
+    page.find("#more-dropdown").click
+    page.find(".item-closeEvent").click
+    page.find("#dialog-holder .btn-primary").click
+
+    expect(page).to have_content("Closed")
```
pmusaraj, line 33: Nitpick, do we need this to be more specific? i.e. page should have an element with a given classname and content "Closed".

spec/system/post_event_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    page.find("#more-dropdown").click
+    page.find(".item-openEvent").click
+    page.find("#dialog-holder .btn-primary").click
+
+    expect(page).to have_content("Public")
```
pmusaraj, line 39: Probably useful to ensure that page no longer has "Closed" here.

pmusaraj (approved):

PR #571, opened 2024-05-21
FIX: delete future holidays on region change

ZogStriP:
When a user changes their holiday region in their profile, we will automatically compute upcoming (6 months) holidays for their new regions. If the events are similar between the two regions, we automatically fix the region. But when they're drastically different, we just leaves the old holiday events be...

It's a somewhat rare event but is annoying enough that it warranted a proper fix.

This ensures we removes any "pre-loaded" holiday events that aren't in the regions the users have selected. That way, no more "fanthom" holidays from the previous region.

Internal ref - t/64785

### DIFF
--- a/jobs/scheduled/create_holiday_events.rb
+++ b/jobs/scheduled/create_holiday_events.rb
@@ -8,11 +8,11 @@ module Jobs
       return if !SiteSetting.calendar_enabled
       return if !SiteSetting.calendar_automatic_holidays_enabled
 
-      topic_id = SiteSetting.holiday_calendar_topic_id.presence
-
-      return if !topic_id
+      return unless topic_id = SiteSetting.holiday_calendar_topic_id.presence
 
       require "holidays" if !defined?(Holidays)
+
+      today = Date.today
 
       regions_and_user_ids = Hash.new { |h, k| h[k] = [] }
 
@@ -51,13 +51,23 @@ module Jobs
           .to_h
 
       # Remove holidays for deactivated/suspended/silenced users
-      CalendarEvent.where(post_id: nil).where.not(user_id: usernames.keys).delete_all
+      CalendarEvent.where(post_id: nil).where.not(user_id: usernames.keys).destroy_all
+
+      # Remove future holidays when users changed their region
+      CalendarEvent
+        .joins(user: :_custom_fields)
+        .where(post_id: nil)
+        .where("start_date > ?", today)
+        .where("user_custom_fields.name = ?", ::DiscourseCalendar::REGION_CUSTOM_FIELD)
+        .where("LENGTH(COALESCE(user_custom_fields.value, '')) > 0")
+        .where("user_custom_fields.value != calendar_events.region")
+        .destroy_all
 
       regions_and_user_ids.each do |region, user_ids|
         DiscourseCalendar::Holiday
           .find_holidays_for(
             region_code: region,
-            start_date: Date.today,
+            start_date: today,
             end_date: 6.months.from_now,
             show_holiday_observed_on_dates: true,
           )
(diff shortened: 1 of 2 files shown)


jobs/scheduled/create_holiday_events.rb
```diff
@@ -8,12 +8,12 @@ def execute(args)
       return if !SiteSetting.calendar_enabled
       return if !SiteSetting.calendar_automatic_holidays_enabled
 
-      topic_id = SiteSetting.holiday_calendar_topic_id.presence
-
-      return if !topic_id
+      return unless topic_id = SiteSetting.holiday_calendar_topic_id.presence
```
ZogStriP, line 11: Just merged the two lines.

jobs/scheduled/create_holiday_events.rb
```diff
@@ -51,13 +51,23 @@ def execute(args)
           .to_h
 
       # Remove holidays for deactivated/suspended/silenced users
-      CalendarEvent.where(post_id: nil).where.not(user_id: usernames.keys).delete_all
+      CalendarEvent.where(post_id: nil).where.not(user_id: usernames.keys).destroy_all
```
ZogStriP, line 54: Changed `delete_all` to `destroy_all`.

jobs/scheduled/create_holiday_events.rb
```diff
@@ -51,13 +51,23 @@ def execute(args)
...
+
+      # Remove future holidays when users changed their region
+      CalendarEvent
+        .joins(user: :_custom_fields)
+        .where(post_id: nil)
+        .where("start_date > ?", today)
+        .where("user_custom_fields.name = ?", ::DiscourseCalendar::REGION_CUSTOM_FIELD)
+        .where("LENGTH(COALESCE(user_custom_fields.value, '')) > 0")
+        .where("user_custom_fields.value != calendar_events.region")
+        .destroy_all
```
ZogStriP, line 64: This query is the gist of the PR. It will automatically remove **future** holiday events that aren't in the region the users have selected in their profile.

jobs/scheduled/create_holiday_events.rb
```diff
@@ -66,14 +76,12 @@ def execute(args)
...
-                  datetime
-                else
-                  holiday[:date]
-                end
+              date = holiday[:date]
+
+              if tz = timezones[user_id]
+                date = holiday[:date].in_time_zone(tz)
+                date = date.change(hour_adjustment) if hour_adjustment
+              end
```
ZogStriP, line 84: Simplified the code a tiny bit. Logic stayed the same.

spec/jobs/scheduled/create_holiday_events_spec.rb
```diff
@@ -31,13 +31,13 @@
     freeze_time Time.zone.local(2019, 8, 1)
     DiscourseCalendar::CreateHolidayEvents.new.execute(nil)
 
-    expect(CalendarEvent.pluck(:region, :description, :start_date, :username)).to match_array(
+    expect(CalendarEvent.pluck(:region, :description, :start_date, :user_id)).to match_array(
```
ZogStriP, line 34: Lots of changes here, but it' because I switched to relying on `username` and instead used `user_id`. Feels safer.

spec/jobs/scheduled/create_holiday_events_spec.rb
```diff
@@ -140,14 +140,56 @@
...
     frenchy.active = false
     frenchy.save!
 
     DiscourseCalendar::CreateHolidayEvents.new.execute(nil)
 
-    expect(CalendarEvent.exists?(username: frenchy.username)).to eq(false)
+    expect(CalendarEvent.exists?(user_id: frenchy.id)).to eq(false)
+  end
+
+  it "cleans up holidays from users who changed their region" do
```
ZogStriP, line 153: This is the spec related to the changes in this PR.

PR #586, opened 2024-06-26
FIX: fix i18n in upcoming-events

chriscpty:
See https://meta.discourse.org/t/broken-no-i18n-in-calendar-plugin-upcoming-events-route/313745

This is a somewhat dirty fix, simply copying the functionality from `discourse-calendar.js` over to `upcoming-events-calendar.js`, but should at least solve the problem.


chriscpty:
resolved with #590

PR #599, opened 2024-08-14
PERF: removed N+1 in availability topic

ZogStriP:
We "recently" did a few changes in the way we use the availability topic

- We used to have 1 topic per year, but have now moved to just using one topic and deleted past posts
- We used to delete past automatic holidays, but we don't anymore since we use them to keep track of #leave days used
- There are more employees covering more regions

All three changes made the query that list all the events to be displayed in the availability topic go 💥

It used to be

- 1 query to list all events from 6 months ago up to 6 months in the future
- 1 query for every events to load the associated user
- 1 query for every events to load the associated's user timezone

So basically, a N^2 + 1 😅

This commit fixes the issue by doing

- 1 query to get all "standalone" events (those associated to a post)
- 1 query to get all the users timezones (just the ones who have an holiday in the current period).
- 1 query to get all the "automatic" events (aka. holidays that are comming from the holidays gem).

So we went from N^2 + 1 down to 3 🚀

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -358,52 +358,81 @@ after_initialize do
   end
 
   add_to_serializer(:post, :calendar_details, include_condition: -> { object.is_first_post? }) do
-    grouped = {}
-    standalones = []
+    start_date = 6.months.ago
 
-    CalendarEvent
-      .where(topic_id: object.topic_id)
-      .where("post_id IS NOT NULL OR start_date >= ?", 6.months.ago)
-      .order(:start_date, :end_date)
-      .each do |event|
-        if event.post_id
-          standalones << {
+    standalone_sql = <<~SQL
+      SELECT post_number, description, start_date, end_date, username, recurrence, timezone
+        FROM calendar_events
+       WHERE topic_id = :topic_id
+         AND post_id IS NOT NULL
+       ORDER BY start_date, end_date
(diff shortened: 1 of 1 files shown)


plugin.rb
```diff
@@ -358,52 +358,81 @@ module ::DiscoursePostEvent
...
+            from: row.start_date,
+            to: row.end_date,
+            username: row.username,
+            recurring: row.recurrence,
+            post_url: Post.url("-", object.topic_id, row.post_number),
+            timezone: row.timezone,
           }
+        end
 
-          user = User.find_by_username(event.username)
```
ZogStriP, line 392: This line was part 1/2 of the issue.

plugin.rb
```diff
@@ -358,52 +358,81 @@ module ::DiscoursePostEvent
...
+            ).select(:user_id),
+        )
+        .where("LENGTH(COALESCE(timezone, '')) > 0")
+        .pluck(:user_id, :timezone)
+        .to_h
 
-          grouped[identifier][:name] << event.description
-          grouped[identifier][:users] << {
-            username: event.username,
-            timezone: user.present? ? user.user_option.timezone : nil,
```
ZogStriP, line 397: This line was part 2/2 of the issue.

PR #603, opened 2024-08-25
FEATURE: Add create event button to editor

poka-IT:
Add a new button "Create an event" on composer

image

Like the deprecated discourse-event plugin did.

We use it in production from this fork for months here on this category: https://forum.monnaie-libre.fr/c/rencontres/21


jjaffeux:
@poka-IT Thanks for the contribution but we are not looking to add this atm. I would suggest you open a topic on meta.discourse.org if you want to discuss this kind of features. If you want to add this for your own forum you could look at using plugin outlets to have this behavior without doing a fork.

PR #604, opened 2024-08-27
DEV: Extract common calendar options

nattsw:
We use the FullCalendar library for this plugin, but certain fixed options are repeated across files. This change consolidates these options, and should introduce consistency.

This also includes a more compatible setting for the locale (referring to example: https://github.com/fullcalendar/fullcalendar/blob/v4.0.0-alpha.3/locales/en-gb.js, "en_GB" -> "en-gb") but it doesn't seem to set the first day correctly. This change also allows for a future improvement which allows correctly* setting of the first day of the week, outside of FullCalendar's default which doesn't seem to reflect correctly.

A future change would include updating us to fullcalendar v6.

### DIFF
--- a/assets/javascripts/discourse/components/upcoming-events-calendar.js
+++ b/assets/javascripts/discourse/components/upcoming-events-calendar.js
@@ -5,8 +5,8 @@ import loadScript from "discourse/lib/load-script";
 import Category from "discourse/models/category";
 import getURL from "discourse-common/lib/get-url";
 import { formatEventName } from "../helpers/format-event-name";
+import fullCalendarDefaultOptions from "../lib/full-calendar-default-options";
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/components/full-calendar-default-options.js
```diff
@@ -0,0 +1,23 @@
+import {
+  getCalendarButtonsText,
+  getCurrentBcp47Locale,
+} from "../lib/calendar-locale";
+import { buildPopover, destroyPopover } from "../lib/popover";
+
+export default function fullCalendarDefaultOptions() {
```
Lhcfl, line 7: would it be better to put it under `assets/javascripts/discourse/lib` folder?

it is placed in the `components` folder, but it is not a component

PR #609, opened 2024-09-16
DEV: Adds ability to see multiday events in sidebar

brrusselburg:
(This PR refers to the [interactivity](https://meta.discourse.org/t/discourse-calendar-and-event/97376#p-473517-integrations-with-other-plugins-7) between the discourse-calendar plugin and the right-sidebar-blocks component.)

As of now, only the first day of an event gets listed in the Upcoming Events sidebar. So if the event is from `10-22-2024 to 10-26-2024,` only the 22nd would have the event on it. On the 23rd, you wouldn't know the event was still happening.

Before:
Screenshot 2024-10-28 at 1 47 38 PM
(Note that this screenshot was taken on October 28th, and the event's start date shows up)

After:
Screenshot 2024-10-28 at 1 36 43 PM
(Taken on the same day, so only the remaining event days show up)

I tested this with both single-day events and multi-day events. We might need to add a limit on the number of results we get back. (If someone throws in a month-long event, it'll get real hairy real quick.) I have some ideas, but I'm not positive of the best place to put that limiter. Ideas welcome!

### DIFF
--- a/assets/javascripts/discourse/components/upcoming-events-list.gjs
+++ b/assets/javascripts/discourse/components/upcoming-events-list.gjs
@@ -107,29 +107,42 @@ export default class UpcomingEventsList extends Component {
   }
 
   groupByMonthAndDay(data) {
+    // this solution worked in the UI, but 5 tests fail
+    // note: does not work if no end date is added to event object
+
     return data.reduce((result, item) => {
       const startDate = moment(item.starts_at);
       const endDate = moment(item.ends_at);
 
-      while (
-        startDate.isSameOrBefore(endDate, "day") &&
-        Object.keys(result).length <= 5 // closer, but still only seeing 1 result in the object.keys...
-      ) {
-        const year = startDate.year();
-        const month = startDate.month() + 1;
-        const day = startDate.date();
+      if (!startDate.isSame(endDate)) {
+        while (startDate.isSameOrBefore(endDate, "day")) {
+          const year = startDate.year();
+          const month = startDate.month() + 1;
+          const day = startDate.date();
+          const monthKey = `${year}-${month}`;
+
+          result[monthKey] = result[monthKey] || {};
+          result[monthKey][day] = result[monthKey][day] || [];
+
+          result[monthKey][day].push(item);
+
+          // Move to the next day
+          startDate.add(1, "day");
+        }
+      } else {
+        const date = new Date(item.starts_at);
+        const year = date.getFullYear();
+        const month = date.getMonth() + 1;
+        const day = date.getDate();
+
         const monthKey = `${year}-${month}`;
 
-        result[monthKey] = result[monthKey] || {};
-        result[monthKey][day] = result[monthKey][day] || [];
+        result[monthKey] = result[monthKey] ?? {};
+        result[monthKey][day] = result[monthKey][day] ?? [];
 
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/components/upcoming-events-list.gjs
```diff
@@ -108,17 +108,41 @@ export default class UpcomingEventsList extends Component {
...
-      const monthKey = `${year}-${month}`;
-
-      result[monthKey] = result[monthKey] ?? {};
-      result[monthKey][day] = result[monthKey][day] ?? [];
-
-      result[monthKey][day].push(item);
+      const startDate = moment(item.starts_at);
+      const endDate = item.ends_at ? moment(item.ends_at) : null;
+
+      if (endDate && !startDate.isSame(endDate)) {
```
jmperez127, line 114: I think there are too many nested blocks. You could return early here instead.

assets/javascripts/discourse/components/upcoming-events-list.gjs
```diff
@@ -108,17 +108,41 @@ export default class UpcomingEventsList extends Component {
...
+          const monthKey = `${year}-${month}`;
+          const today = moment();
+
+          // For current events, only push upcoming days
+          if (startDate.isAfter(today)) {
+            result[monthKey] = result[monthKey] || {};
+            result[monthKey][day] = result[monthKey][day] || [];
+
+            result[monthKey][day].push(item);
+          }
```
jmperez127, line 128: This part is essentially repeated here and on the following else block. You could create a helper function to DRY things up, and just keep the if statement here.

Something like:

```javascript
function addToResult(date, item, result) {
  const year = date.year();
  const month = date.month() + 1;
  const day = date.date();
  const monthKey = `${year}-${month}`;

  result[monthKey] = result[monthKey] || {};
  result[monthKey][day] = result[monthKey][day] || [];
  result[monthKey][day].push(item);
}

//....
// then you can do something like this:

return data.reduce((result, item) => {
  const startDate = moment(item.starts_at);
  const endDate = item.ends_at ? moment(item.ends_at) : null;

  if (!endDate || startDate.isSame(endDate, "day")) {
    addToResult(startDate, item, result);
    return result;
  }

  while (startDate.isSameOrBefore(endDate, "day")) {
    if (startDate.isAfter(today)) {
      addToResult(startDate, item, result);
    }
    startDate.add(1, "day");
  }

  return result;
}, [])
```

jmperez127 (changes requested):

assets/javascripts/discourse/components/upcoming-events-list.gjs
```diff
@@ -17,6 +17,17 @@ export const DEFAULT_TIME_FORMAT = "LT";
 const DEFAULT_UPCOMING_DAYS = 180;
 const DEFAULT_COUNT = 8;
 
+function addToResult(date, item, result) {
+  const year = date.year();
+  const month = date.month() + 1;
+  const day = date.date();
+  const monthKey = `${year}-${month}`;
```
jmperez127, line 24: I think you could simply do `const monthKey = date.format("YYYY-MM");` and remove the year and month constants

jmperez127 (approved):

PR #611, opened 2024-10-03
FIX: shows category calendar on hot/latest

jjaffeux:
The previous URL parsing code was not resilient to these paths, we now use the router to recognize these paths and extract the params we need.


spec/system/category_calendar_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+describe "Post event", type: :system do
```
jjaffeux, line 3:
```suggestion
describe "Category calendar", type: :system do
```

PR #613, opened 2024-10-09
UX: fix event builder i18n

renato:
Replaces hardcoded strings on the event builder modal by existing i18n strings.


assets/javascripts/discourse/components/modal/post-event-builder.js
```diff
@@ -5,6 +5,7 @@ import { inject as service } from "@ember/service";
 import { extractError } from "discourse/lib/ajax-error";
 import { cook } from "discourse/lib/text";
 import Group from "discourse/models/group";
+import I18n from "I18n";
```
CvX, line 8: the new/recommended module path is now:

```suggestion
import I18n from "discourse-i18n";
```

CvX (approved):

PR #623, opened 2024-10-22
FIX: correctly render html of event title

jjaffeux:
This has been lost during the move to glimmer.

### DIFF
--- a/assets/javascripts/discourse/components/discourse-post-event/index.gjs
+++ b/assets/javascripts/discourse/components/discourse-post-event/index.gjs
@@ -1,13 +1,11 @@
 import Component from "@glimmer/component";
 import { hash } from "@ember/helper";
 import { service } from "@ember/service";
-import { htmlSafe } from "@ember/template";
 import { modifier } from "ember-modifier";
 import PluginOutlet from "discourse/components/plugin-outlet";
 import concatClass from "discourse/helpers/concat-class";
+import replaceEmoji from "discourse/helpers/replace-emoji";
 import routeAction from "discourse/helpers/route-action";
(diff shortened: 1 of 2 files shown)


pmusaraj (approved):

assets/javascripts/discourse/components/discourse-post-event/index.gjs
```diff
@@ -53,8 +54,11 @@ export default class DiscoursePostEvent extends Component {
...
 
   get eventName() {
-    return emojiUnescape(
-      escapeExpression(this.args.event.name) || this.args.event.post.topic.title
+    return htmlSafe(
+      emojiUnescape(
+        escapeExpression(this.args.event.name) ||
+          this.args.event.post.topic.title
+      )
     );
```
davidtaylorhq, line 62: We need to escape `topic.title` as well.

If we want to keep this all manual, then I think we should split it up like:
```suggestion
    const unsafeRawTitle = this.args.event.name || this.args.event.post.topic.title;
    const escapedTitle = escapeExpression(unsafeRawTitle);
    const escapedTitleWithEmoji = emojiUnescape(escapedTitle);
    return htmlSafe(escapedTitleWithEmoji);
```

---

Or, even better, we can do the emoji escaping using our hbs helper which has all the escaping built-in.

So make the getter a simple thing like:

```
get eventName(){
  return this.args.event.name || this.args.event.post.topic.title;
}
```

and then import/use the helper:
...
```gjs
import replaceEmoji from "discourse/helpers/replace-emoji";
...
<template>
  {{replaceEmoji this.eventName}}
</template>
```

davidtaylorhq (approved):

PR #632, opened 2024-10-24
FEATURE: keeps going invitees for recurring events

jjaffeux:
Prior to this fix when computing the next event date we were resetting the status of every invitees. The status will now only be reset for non going users.

### DIFF
--- a/app/models/discourse_post_event/event.rb
+++ b/app/models/discourse_post_event/event.rb
@@ -53,12 +53,14 @@ module DiscoursePostEvent
         end
       end
 
-      publish_update!
-      invitees.update_all(status: nil, notified: false)
+      invitees.where.not(status: Invitee.statuses[:going]).update_all(status: nil, notified: false)
+
       if !next_dates[:rescheduled]
         notify_invitees!
         notify_missing_invitees!
       end
+
+      publish_update!
     end
 
     def set_topic_bump
--- a/app/models/discourse_post_event/event_date.rb
+++ b/app/models/discourse_post_event/event_date.rb
@@ -41,11 +41,6 @@ module DiscoursePostEvent
       end
     end
 
-    after_commit :reset_invitees_status, on: %i[create]
-    def reset_invitees_status
-      self.event.invitees.update_all(status: nil)
-    end
-
     def started?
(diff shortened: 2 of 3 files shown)


app/models/discourse_post_event/event.rb
```diff
@@ -53,12 +53,14 @@ def set_next_date
...
-      publish_update!
-      invitees.update_all(status: nil, notified: false)
+      invitees.where.not(status: Invitee.statuses[:going]).update_all(status: nil, notified: false)
+
       if !next_dates[:rescheduled]
         notify_invitees!
         notify_missing_invitees!
       end
+
+      publish_update!
```
jjaffeux, line 63: Not really related, but I think we should publish the update once we are done, and not randomly in the middle of the sequence

app/models/discourse_post_event/event.rb
```diff
@@ -53,12 +53,14 @@ def set_next_date
         end
       end
 
-      publish_update!
-      invitees.update_all(status: nil, notified: false)
+      invitees.where.not(status: Invitee.statuses[:going]).update_all(status: nil, notified: false)
```
jjaffeux, line 56: This is the main change. Instead of selectioning every invitees we filter out the going invitees

app/models/discourse_post_event/event_date.rb
```diff
@@ -41,11 +41,6 @@ def upsert_topic_custom_field
       end
     end
 
-    after_commit :reset_invitees_status, on: %i[create]
```
jjaffeux, line 44: the only place where we call event_dates.create and trigger this hook, is in the event hook and we manually call the set_date which has the same behavior we just have modified in this commit. It was essentially doing two times the same work.

nbianca (approved):

app/models/discourse_post_event/event.rb
nbianca: I am wondering if there should be a "going to this event" or "going to all event options" like other calendars have.

jjaffeux: maybe, but for now we will assume it's by default "going to this and all future events"

PR #649, opened 2024-11-25
FEATURE: Allow users to set a custom title to sidebar calendar

brrusselburg:
This allows users to change the title of the calendar sidebar according to the category. The example use-case is "Upcoming Releases" in the "Product News" category.

Default:
Screenshot 2024-11-27 at 1 50 45 PM

Custom:
Screenshot 2024-11-27 at 1 51 30 PM

Setting:
Screenshot 2024-11-27 at 1 51 15 PM

Tests will be added into the Right Sidebar Blocks component, as this feature only works in conjunction with specific settings in that component.

### DIFF
--- a/assets/javascripts/discourse/components/upcoming-events-list.gjs
+++ b/assets/javascripts/discourse/components/upcoming-events-list.gjs
@@ -86,9 +86,7 @@ export default class UpcomingEventsList extends Component {
       );
       const custom_title = customTitleLookup?.custom_title;
 
-      return I18n.t("discourse_post_event.upcoming_events_list.custom_title", {
-        upcoming_events_title: custom_title,
-      });
+      return custom_title;
     } else {
       return I18n.t("discourse_post_event.upcoming_events_list.title");
     }
--- a/assets/javascripts/discourse/components/upcoming-events-list.gjs
+++ b/assets/javascripts/discourse/components/upcoming-events-list.gjs
@@ -84,9 +84,7 @@ export default class UpcomingEventsList extends Component {
       const customTitleLookup = titleMap.find(
         (o) => o.category_slug === categorySlug
       );
-      const custom_title = customTitleLookup?.custom_title;
-
-      return custom_title;
+      return customTitleLookup?.custom_title;
     } else {
       return I18n.t("discourse_post_event.upcoming_events_list.title");
     }
--- a/assets/javascripts/discourse/components/upcoming-events-list.gjs
+++ b/assets/javascripts/discourse/components/upcoming-events-list.gjs
@@ -26,6 +26,21 @@ function addToResult(date, item, result) {
   result[monthKey][day].push(item);
 }
 
+// function title() {
+//   const titleMap = JSON.parse(siteSettings);
+//   const categorySlug = this.args.params?.categorySlug;
+
+//   const obj = titleMap.find((o) => o.category_slug === categorySlug);
+//   const customTitle = customTitleObj
+//     ? customTitleObj.custom_title
+//     : "Upcoming Events";
+//   // technically the default value in the setting, but wouldn't hurt to have it here as well?
+
+//   return I18n.t("discourse_post_event.upcoming_events_list.title", {
+//     upcoming_events_title: customTitle,
+//   });
+// }
+
 export default class UpcomingEventsList extends Component {
   @service appEvents;
   @service siteSettings;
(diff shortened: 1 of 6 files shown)


renato:
I think this behavior might be entirely tested within the existing integration test we have here, no? What specifics of the right sidebar blocks we need to cover?

If it's just the category we can mock different categories with the RouterStub for different tests

brrusselburg:
> I think this behavior might be entirely tested within the existing integration test we have here, no? What specifics of the right sidebar blocks we need to cover?
> 
> If it's just the category from the route, we can mock different categories with the RouterStub for different tests

The default title is covered, but I'm not sure I see how a custom path would be 🤔 And yeah, the issue we had the other day was in getting the context of a specific category into this test.

assets/javascripts/discourse/components/upcoming-events-list.gjs
```diff
@@ -68,6 +67,36 @@ export default class UpcomingEventsList extends Component {
     return this.router.currentRoute.attributes?.category?.id;
   }
 
+  get title() {
+    const categorySlug = this.router.currentRoute.attributes?.category?.slug;
+    const titleSetting = this.siteSettings?.map_events_title;
```
CvX, line 72: siteSettings is always present

```suggestion
    const titleSetting = this.siteSettings.map_events_title;
```

assets/javascripts/discourse/components/upcoming-events-list.gjs
```diff
@@ -68,6 +67,36 @@ export default class UpcomingEventsList extends Component {
...
+        (o) => o.category_slug === categorySlug
+      );
+      const custom_title =
+        typeof customTitleLookup === "undefined"
+          ? null
+          : customTitleLookup.custom_title;
+
+      return I18n.t("discourse_post_event.upcoming_events_list.custom_title", {
+        upcoming_events_title: custom_title,
+      });
```
CvX, line 94: golf + case

```suggestion
      const customTitle = customTitleLookup?.custom_title;

      return I18n.t("discourse_post_event.upcoming_events_list.custom_title", {
        upcoming_events_title: customTitle,
      });
```

config/locales/client.en.yml
```diff
@@ -373,6 +373,7 @@ en:
         starts_at: "Starts at"
       upcoming_events_list:
         title: "Upcoming events"
+        custom_title: "%{upcoming_events_title}"
```
CvX, line 376: do we need a i18n string for this?

brrusselburg: Probably not? I split them to make it easier for me to follow while I worked on the solution.

CvX: yeah, I'd drop it

assets/javascripts/discourse/components/upcoming-events-list.gjs
```diff
@@ -68,6 +67,31 @@ export default class UpcomingEventsList extends Component {
...
+    );
+
+    if (categories.includes(categorySlug)) {
+      const titleMap = JSON.parse(titleSetting);
+      const customTitleLookup = titleMap.find(
+        (o) => o.category_slug === categorySlug
+      );
+      const custom_title = customTitleLookup?.custom_title;
+
+      return custom_title;
```
CvX, line 89: shorter and with no underscore_case variable:

```suggestion
      return customTitleLookup?.custom_title;
```

test/javascripts/integration/components/upcoming-events-list-test.gjs
```diff
@@ -163,6 +163,41 @@ module("Integration | Component | upcoming-events-list", function (hooks) {
...
+      '[{"category_slug": "announcements", "custom_title": "Upcoming Announcements"}]';
+
+    await render(<template><UpcomingEventsList /></template>);
+    this.appEvents.trigger("page:changed", { url: "/c/announcements" });
+
+    assert.strictEqual(
+      query(".upcoming-events-list__heading").innerText,
+      "Upcoming Announcements",
+      "it sets 'Upcoming Announcements' as the title in 'c/announcements'"
+    );
```
CvX, line 181: qunit-dom assertions are the only recommended way to test text/html/existence/etc going forward

```suggestion
    assert
      .dom(".upcoming-events-list__heading")
      .hasText(
        "Upcoming Announcements",
        "sets 'Upcoming Announcements' as the title in 'c/announcements'"
      );
```

brrusselburg:
> Sorry @brrusselburg ...
> 
> Is there a reason why we're not using text replacement feature here? https://meta.discourse.org/t/customize-text-in-discourse/36092
> 
> http://localhost:4200/admin/customize/site_texts?q=discourse_post_event.upcoming_events_list.title Screenshot 2024-12-03 at 11 18 48 AM

Yep. That was originally being used, but it doesn't account for different titles per route.

test/javascripts/integration/components/upcoming-events-list-test.gjs
brrusselburg: good to know, thank you!

assets/javascripts/discourse/components/upcoming-events-list.gjs
brrusselburg: ope, totally missed that. thank you.

CvX (approved):

PR #641, opened 2024-11-01
UX: Events UI Edits

jordanvidrine:
This PR restyles the current events post UI:

- Minimize size of invitee avatars
- move CTA buttons to bottom
- remove invitee header & show more button (show more is in ellipses drop down)
- restyle border
- restyle month & date size & spacing
- increase font size for title
- add "0 going" placeholder to prevent jumpiness when invitees are rendered

Related PR -> https://github.com/discourse/discourse-calendar/pull/654

**Before**
image

**After**
image

PR #654, opened 2024-11-29
UX: Tc right sidebar changes

jordanvidrine:
This PR brings the right sidebar injected component up to date with the topic-post event rendering. (when the right sidebar blocks theme component is installed & active)

Related PR -> https://github.com/discourse/discourse-calendar/pull/641

**After**
image

**Before**
image


brrusselburg:
👨‍🍳 WOW, looks so good!! 😭 

Have you tested how the new design will work if certain elements are removed from the time syntax? For ex, `timeFormat` or `monthFormat` keys left blank in the settings.

Screenshot 2024-11-29 at 6 03 33 PM

jordanvidrine:
I have not. But my gut says. "Dear lord why is this even an allowed setting?!"

PR #655, opened 2024-11-29
FIX: Render component on all routes

jordanvidrine:
This PR fixes the rendering of this component on all routes (when defined in the rightsidebar blocks component).

It was rendering a large empty space when *not* in an events category, but will now render all site events when on a non-category specific route. When on a category, it will only render the events within that category.

**Before**
image

**After**
image


brrusselburg:
Here's what I see when I pull this down.

My local env was already set up to have two categories with the Upcoming Events List. Both of them looked good. I added a new category to the Right Sidebar Blocks settings. The expectation was for me to see a blank right sidebar, since I hadn't yet included the `staff` category into the Calendar plugin settings.

Setup:
Screenshot 2024-11-29 at 6 06 21 PM
user-attachments/assets/851a9531-3074-400a-bee1-bd425d84367b">

Result on `/c/staff`:
Screenshot 2024-11-29 at 6 10 20 PM

jordanvidrine:
This is a problem, but it's a problem caused by being able to set the right sidebar blocks component on a category that does not include events. This is more of a user problem than a component problem. 

If a user wants to render it in a category that does not have events, that is up to them I guess?

brrusselburg (approved):

mow:
Hi there, indeed that this issue is a bit more a business decision rather than technical. In my opinion, if users are able to view all events from the "upcoming event" left sidebar menu link, this widget should provide the same. In order not to break existing behavior and provide the feature I quickly created https://github.com/mow/discourse-calendar/pull/1 (from the inline editor), if this seems logical to you I'll create the PR, thank you & have a nice day!

PR #656, opened 2024-12-02
DEV: Update modifyClass calls to native class syntax

davidtaylorhq:


CvX (approved):
This is also included in my PR but we can merge this for cleaner history 😉

PR #663, opened 2024-12-05
UX: Render invitees count

jordanvidrine:
This PR adds a "going" count back into the events UI. The counts were removed previously to minimize the amount of information, but "going" should have remained.

**After**
CleanShot 2024-12-05 at 15 48 00@2x

**Before**
CleanShot 2024-12-05 at 15 48 48@2x

### DIFF
--- a/assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
+++ b/assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
@@ -1,7 +1,6 @@
 import Component from "@glimmer/component";
 import { action } from "@ember/object";
 import { service } from "@ember/service";
-import DButton from "discourse/components/d-button";
 import icon from "discourse-common/helpers/d-icon";
 import i18n from "discourse-common/helpers/i18n";
 import PostEventInvitees from "../modal/post-event-invitees";
--- a/assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
+++ b/assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
@@ -1,6 +1,7 @@
 import Component from "@glimmer/component";
 import { action } from "@ember/object";
 import { service } from "@ember/service";
+import DButton from "discourse/components/d-button";
 import icon from "discourse-common/helpers/d-icon";
 import i18n from "discourse-common/helpers/i18n";
 import PostEventInvitees from "../modal/post-event-invitees";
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
```diff
@@ -22,57 +21,31 @@ export default class DiscoursePostEventInvitees extends Component {
...
-              {{#each this.statsInfo as |info|}}
-                <span class={{info.class}}>{{info.label}}</span>
-              {{/each}}
-            </div>
-
-            <DButton
-              class="show-all btn-small"
-              @label="discourse_post_event.show_all"
-              @action={{this.showAllInvitees}}
-            />
```
CvX, line 72: If this removal was intentional then `showAllInvitees` method should be deleted too

assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
```diff
@@ -22,57 +21,31 @@ export default class DiscoursePostEventInvitees extends Component {
...
-
-    visibleStats.forEach((button) => {
-      const localeKey = button.replace(" ", "_");
-      if (button === "not_going") {
-        button = "notGoing";
-      }
-
-      const count = this.args.event.stats[button] || 0;
+  get hasAttendees() {
+    return this.args.event.stats["going"] > 0;
```
CvX, line 25: btw, in all three places you can write it like this:

```suggestion
    return this.args.event.stats.going > 0;
```

assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
jordanvidrine: Yes this was intentional, this button is now avail in the drop down ellipsis menu.

jordanvidrine: will remove the method.

CvX: now the component can be reduced to a template-only component, both service injections are unused and the properties could be inlined in the template 😃

jordanvidrine: Would this still remain a `gjs` file? and only include <template> while still using the imports like:

```
import icon from "discourse-common/helpers/d-icon";
import i18n from "discourse-common/helpers/i18n";
import Invitee from "./invitee";
```

CvX: yup - still a gjs but without a js class:

```gjs
const DiscoursePostEventInvitees = <template>
  ...
</template>

export default DiscoursePostEventInvitees;
```

(btw, the new i18n import is: `import { i18n } from "discourse-i18n";`)

PR #666, opened 2024-12-10
FIX: ensures bulk-invite is working properly

jjaffeux:
There was two bugs:
- naming of properties were incorrect s/is_private/isPrivate and s/is_public/isPublic
- a previous refactoring has used `=` for setter when it's an `EmberObject` and `set` should be used

This commit also adds a spec and page objects to ensure this modal is working as expected.


spec/system/post_event_spec.rb
```diff
@@ -92,4 +94,29 @@
...
+    fab!(:invitable_user_2) { Fabricate(:user) }
+
+    it "can invite users to an event" do
+      visit(post.topic.url)
+
+      post_event_page.open_bulk_invite_modal
+      bulk_invite_modal_page.set_invitee_at_row(invitable_user_1.username, "going", 1)
+      bulk_invite_modal_page.add_invitee
+      bulk_invite_modal_page.set_invitee_at_row(invitable_user_2.username, "not_going", 2)
+      bulk_invite_modal_page.send_invites
```
Grubba27, line 117: would something like this work?
```suggestion
      bulk_invite_modal_page
        .set_invitee_at_row(invitable_user_1.username, "going", 1)
        .add_invitee
        .set_invitee_at_row(invitable_user_2.username, "not_going", 2)
        .send_invites
```

jjaffeux: yes but Im mixed on this being really better

jjaffeux: Ok I did, because I don't know... let's say it's better :+1:

ZogStriP (approved):

spec/system/post_event_spec.rb
ZogStriP: I ❤️ the chained version slightly more than the procedural one 👍

Grubba27: I feel that has less _noise_

PR #671, opened 2024-12-20
FIX: correctly apply local dates on event dates

jjaffeux:
The next ensures `this.htmlDates` has correctly been set and a render is on going which is going to be awaited by the schedule render. Before this fix `querySelectorAll` could return an empty nodes list as the dates were not rendered yet.

### DIFF
--- a/assets/javascripts/discourse/components/discourse-post-event/dates.gjs
+++ b/assets/javascripts/discourse/components/discourse-post-event/dates.gjs
@@ -2,7 +2,7 @@ import Component from "@glimmer/component";
 import { tracked } from "@glimmer/tracking";
 import { action } from "@ember/object";
 import didInsert from "@ember/render-modifiers/modifiers/did-insert";
-import { schedule, next } from "@ember/runloop";
+import { next, schedule } from "@ember/runloop";
 import { service } from "@ember/service";
 import { htmlSafe } from "@ember/template";
 import { applyLocalDates } from "discourse/lib/local-dates";
--- a/assets/javascripts/discourse/components/discourse-post-event/dates.gjs
+++ b/assets/javascripts/discourse/components/discourse-post-event/dates.gjs
@@ -2,7 +2,7 @@ import Component from "@glimmer/component";
 import { tracked } from "@glimmer/tracking";
 import { action } from "@ember/object";
 import didInsert from "@ember/render-modifiers/modifiers/did-insert";
-import { schedule } from "@ember/runloop";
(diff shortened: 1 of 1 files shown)


ZogStriP (approved):

assets/javascripts/discourse/components/discourse-post-event/dates.gjs
```diff
@@ -59,13 +59,15 @@ export default class DiscoursePostEventDates extends Component {
...
-        );
+      next(() => {
+        schedule("afterRender", () => {
+          applyLocalDates(
+            element.querySelectorAll(
+              `[data-post-id="${this.args.event.id}"] .discourse-local-date`
+            ),
+            this.siteSettings
+          );
+        });
```
CvX, line 70: 1. `schedule("afterRender")` inside `next()` is extraneous - `next` already schedules a separate run loop, so specifying that this code needs to run in the "afterRender" part of it shouldn't change anything
2. you might need to add `isDestroying || isDestroyed` check in there so that the async execution doesn't bleed between tests

jjaffeux: > 1. `schedule("afterRender")` inside `next()` is extraneous - `next` already schedules a separate run loop, so specifying that this code needs to run in the "afterRender" part of it shouldn't change anything

And yet I have 100% repro without, and 0% with this next

jjaffeux: > 2\. you might need to add `isDestroying || isDestroyed` check in there so that the async execution doesn't bleed between tests

:+1:

CvX: > 0% with this next

Just to confirm: I'm suggesting using `next` w/o `schedule`, not the other way around. Is `schedule` critical to the fix?

jjaffeux: > 1. `schedule("afterRender")` inside `next()` is extraneous - `next` already schedules a separate run loop, so specifying that this code needs to run in the "afterRender" part of it shouldn't change anything

Given it fixes my immediate issue I will merge it, but happy to revert if we have a better solution, or if it ends up not fixing it

jjaffeux: > > 0% with this next
> 
> Just to confirm: I'm suggesting using `next` w/o `schedule`, not the other way around. Is `schedule` critical to the fix?

AH I had missed this, will try :+1:

PR #672, opened 2024-12-24
UX: Remove going indicator for standalone events

jordanvidrine:
This PR removes the "0 Going" indicator if an event is standalone. Because users cannot mark they are "going" to a standalone event, the "0 Going" was always present, which is not ideal.

I also added a small line of CSS to always add padding to the bottom of the last section equal to the to section in the event post component.

**Before**
CleanShot 2024-12-24 at 08 01 45@2x

**After**
CleanShot 2024-12-24 at 08 01 27@2x

### DIFF
--- a/assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
+++ b/assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
@@ -41,12 +41,14 @@ export default class DiscoursePostEventInvitees extends Component {
           </div>
         </section>
       {{else}}
-        <section class="event__section event-invitees no-rsvp">
-          <p class="no-rsvp-description">{{i18n
-              "discourse_post_event.models.invitee.status.going_count.other"
-              count="0"
-            }}</p>
-        </section>
+        {{#unless @event.isStandalone}}
+          <section class="event__section event-invitees no-rsvp">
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
```diff
@@ -41,12 +41,14 @@ export default class DiscoursePostEventInvitees extends Component {
...
-          <p class="no-rsvp-description">{{i18n
-              "discourse_post_event.models.invitee.status.going_count.other"
-              count="0"
-            }}</p>
-        </section>
+        {{#unless @event.isStandalone}}
+          <section class="event__section event-invitees no-rsvp">
+            <p class="no-rsvp-description">{{i18n
+                "discourse_post_event.models.invitee.status.going_count.other"
+                count="0"
```
Grubba27, line 48: Is this `count="0"` correct?

Grubba27 (approved):
looks 🔥

assets/javascripts/discourse/components/discourse-post-event/invitees.gjs
jordanvidrine: That is correct, this only appears when there are no respondents.

PR #679, opened 2025-01-14
UX: Do not show participants in admin drop down menu for standalone events

jordanvidrine:
This PR adds logic in order to *not* show the "show participants" button in the `more` dropdown on standalone events.

**After**
CleanShot 2025-01-14 at 15 03 06@2x

**Before**
CleanShot 2025-01-14 at 15 03 52@2x


jordanvidrine:
> LGTM overall, but can we add a system spec here to prevent regression? Also a screenshot in the PR would be awesome for easy reference :)

Keegan, this test passed locally for me but not on github. Would you mind taking a look and see if you spot something wrong? This is the first time for me writing a ruby spec test.

PR #681, opened 2025-01-20
DEV: schedule is un-necessary

jjaffeux:
As per CvX:

> schedule("afterRender") inside next() is extraneous - next already schedules a separate run loop, so specifying that this code needs to run in the "afterRender" part of it shouldn't change anything


jjaffeux:
Seems like it's enough @CvX , feel free to merge directly

CvX (approved):

PR #683, opened 2025-01-23
DEV: Add DiscourseEvent triggers to Invitee creation and update

jdmartinez1062:
**Description**

We need to add these Event triggers as part of changes made to the discourse-livestream plugin to allow the user chat allow list setting to work properly, check https://github.com/discourse/discourse-livestream/pull/43 for more context


featheredtoast (changes requested):
Can we rename the trigger to something that distinguishes that the events are coming from this plugin? `discourse_calendar_post_event....`

Also was the addition of the discourse post event user extension expected in this batch of changes?

featheredtoast (approved):

PR #689, opened 2025-02-10
Fix for inconsistent first-day-of-the-week display.

communiteq:
There are three places in this plugin where FullCalendar is set up.

In one of these three the parameter firstDay: 1 is passed, in the other two it's not. This leads to an inconsistency where the Upcoming Events calendar begins on Sunday, whereas the topic calendars start on Monday. (See https://meta.discourse.org/t/choose-first-day-of-the-week-on-upcoming-events-calendar/307785/5)

Ideally this would be taken from the users (browser) preferences, but since [all other places in core](https://github.com/search?q=repo%3Adiscourse%2Fdiscourse%20firstDay&type=code) hard code it to firstDay: 1 as well, this small patch at least adds some consistency.

### DIFF
--- a/assets/javascripts/discourse/components/upcoming-events-calendar.js
+++ b/assets/javascripts/discourse/components/upcoming-events-calendar.js
@@ -46,6 +46,7 @@ export default class UpcomingEventsCalendar extends Component {
     this._loadCalendar().then(() => {
       const fullCalendar = new window.FullCalendar.Calendar(calendarNode, {
         ...fullCalendarDefaultOptions(),
+        firstDay: 1,
         height: "auto",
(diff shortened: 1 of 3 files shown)


pmusaraj:
Makes sense to me @communiteq. @jjaffeux you've been working on this plugin recently, WDYT?

Also, failing specs are legitimate, I believe.

jjaffeux (approved):
:+1:

jjaffeux:
> Makes sense to me @communiteq. @jjaffeux you've been working on this plugin recently, WDYT?
> 
> Also, failing specs are legitimate, I believe.

Yes fine, I fixed the tests and will merge it.

PR #693, opened 2025-02-21
FEATURE: add plugin outlet in modal/post-event-builder

MarineBaron:
This add PluginOutlet in modal/post-event-builder component.

We need this to override modal with our customs fields which are not all input text.

Thanks !


pmusaraj:
Seems fine, thoughts @jjaffeux?

PR #698, opened 2025-03-10
PERF: Fix N+1 problem on `DiscoursePostEvent::EventsController#index`

tgxworld:


spec/requests/events_controller_spec.rb
```diff
@@ -9,6 +9,28 @@ module DiscoursePostEvent
...
+
+        event_2 = Fabricate(:event, original_starts_at: 2.days.from_now)
+        event_3 = Fabricate(:event, original_starts_at: 3.days.from_now)
+
+        new_queries = track_sql_queries { get "/discourse-post-event/events.json" }
+
+        expect(response.status).to eq(200)
+        expect(response.parsed_body["events"].length).to eq(3)
+
+        # TODO: Remove the 4 queries due to another N+1 problem.
```
tgxworld, line 29: @dbattersby Your fix should end up removing the need for `.count - 4` in the subsequent line.

dbattersby: Okay yep sounds good. We can tweak that after I fix the other issue.

PR #704, opened 2025-03-14
DEV: Update linting

CvX:


assets/stylesheets/colors.scss
```diff
@@ -1,14 +1,16 @@
+@use "sass:color";
```
CvX, line 1: @davidtaylorhq this fixes the issue from the previous PR

I also added a "compile color scheme stylesheet" step to the scss deprecation checker

davidtaylorhq (approved):

PR #706, opened 2025-03-18
DEV: Allow the 'UpcomingEventsCalendar' Component to be used outside of the 'PostEventUpcomingEventsIndexRoute' route

jdmartinez1062:


assets/javascripts/discourse/services/discourse-post-event-service.js
```diff
@@ -0,0 +1,14 @@
+import Service, { service } from "@ember/service";
+
+export default class DiscoursePostEventService extends Service {
+  @service siteSettings;
+  @service discoursePostEventApi;
+
+  async fetchEvents(params = {}) {
```
brrusselburg, line 7: should we test this?

assets/javascripts/discourse/lib/add-recurrent-events.js
```diff
@@ -1,18 +1,30 @@
...
-          name: event.name,
-          post: event.post,
-          category_id: event.categoryId,
-          starts_at: upcomingDate.starts_at,
-          ends_at: upcomingDate.ends_at,
-        })
-      ) || [];
+export default async function addRecurrentEvents(eventsPromise) {
+  try {
+    const events = await eventsPromise;
```
markvanlan, line 5: I think it's a weird to do the `await` in here.. It feels like this function should expect an array instead of a promise

markvanlan (approved):

PR #708, opened 2025-03-21
FEATURE: adds support for event preview in rich editor

jjaffeux:


jjaffeux:
Screenshot 2025-03-21 at 17 45 56

This will display as the composer preview does in markdown mode atm.

test/javascripts/integration/components/rich-editor-extension-test.js
```diff
@@ -0,0 +1,27 @@
+import { module, test } from "qunit";
+import { setupRenderingTest } from "discourse/tests/helpers/component-test";
+import { testMarkdown } from "discourse/tests/helpers/rich-editor-helper";
+
+module("Integration | Component | rich-editor-extension", function (hooks) {
+  setupRenderingTest(hooks);
+
+  const testCases = {
+    emoji: [
```
renato, line 9:
```suggestion
    event: [
```

jjaffeux: oops

assets/javascripts/discourse/pre-initializers/rich-editor.extension.js
```diff
@@ -0,0 +1,136 @@
...
+  allowedGroups: { default: null },
+};
+
+/** @type {RichEditorExtension} */
+const extension = {
+  nodeSpec: {
+    event: {
+      attrs: EVENT_ATTRIBUTES,
+      group: "block",
+      content: "block+",
```
renato, line 23: we only need to set a `content` when we want this node to have children, this would define what children are allowed
```suggestion
```

assets/javascripts/discourse/pre-initializers/rich-editor.extension.js
```diff
@@ -0,0 +1,136 @@
...
+
+  parse: {
+    wrap_bbcode(state, token) {
+      if (token.tag === "div") {
+        if (token.nesting === -1) {
+          state.closeNode();
+          return true;
+        }
+
+        if (token.nesting === 1) {
```
renato, line 85: this feels too broad, maybe...
```suggestion
      if (token.tag === "div") {
        if (token.nesting === -1 && state.top().type.name === "event") {
          state.closeNode();
          return true;
        }

        if (token.nesting === 1 && token.attrGet("class") === "discourse-post-event") {
```
this also makes me wonder if we couldn't use a different token on markdown-it, we're opening/closing this token but we don't expect any content inside it, do we?

jjaffeux: Ah yes I was trying to do this sooner but forgot, thanks

jjaffeux:
@renato I fixed tests

assets/javascripts/discourse/pre-initializers/rich-editor.extension.js
```diff
@@ -0,0 +1,140 @@
...
+    event(state, node) {
+      let bbcode = "[event";
+
+      Object.entries(node.attrs).forEach(([key, value]) => {
+        if (value !== null) {
+          bbcode += ` ${key}="${value}"`;
+        }
+      });
+
+      bbcode += "]\n[/event]";
```
renato, line 127: I believe we need a trailing `\n`.

Maybe a test with content before and after the bbcode may help catch this.

assets/javascripts/discourse/pre-initializers/rich-editor.extension.js
```diff
@@ -0,0 +1,140 @@
...
+              acc[`data-${key.replace(/([A-Z])/g, "-$1").toLowerCase()}`] =
+                value;
+            }
+            return acc;
+          },
+          {}
+        );
+
+        const attrs = {
+          class: "discourse-post-event-preview",
```
renato, line 48: we're parsing `div.discourse-post-event` but creating `div.discourse-post-event-preview`, so we can't copy/cut/paste the node correctly. ideally we keep both consistent – and hopefully consistent with cooked but that's not always possible without bigger changes – so we can copy/paste more freely

renato:
I think the PR title can be a `FEATURE:`, wdyt?

assets/javascripts/discourse/pre-initializers/rich-editor.extension.js
renato, line 1: Just curious, is the name `rich-editor.extension.js` intentional?

assets/javascripts/discourse/pre-initializers/rich-editor.extension.js
```diff
@@ -0,0 +1,140 @@
...
+          getAttrs(dom) {
+            return { ...dom.dataset };
+          },
+        },
+      ],
+      toDOM(node) {
+        const dataAttrs = Object.entries(node.attrs).reduce(
+          (acc, [key, value]) => {
+            if (value !== null) {
+              acc[`data-${key.replace(/([A-Z])/g, "-$1").toLowerCase()}`] =
```
renato, line 39: I think we can avoid this camel<>kebab conversion if we use a different strategy, but creating the element itself instead of returning the array notation.

Something like...

```javascript
const element = document.createElement("div");
for (const [key, value] of Object.entries(node.attrs)) {
  if (value !== null) {
    element.dataset[key] = value;
  }
}
```

Then you can `element.appendChild` and `return element` in the end.

renato (approved):

renato:
@jjaffeux I'm merging it (after having applied some changes)

jjaffeux:
ah yes sorry was going to address these next week, but thanks

PR #716, opened 2025-04-08
FIX: remove double quotes `"` character when building the event's mar…

ZogStriP:
…kdown

This is not the best UX as user might enter an event name with double quotes and they'll be deleted once they click save. But at least it won't break their event because the double quote breaks the BBCode/Markdown.

A proper fix would be to manipulate an AST instead of using regular expressions on a a string.

Meta - https://meta.discourse.org/t/-/360010

### DIFF
--- a/test/javascripts/lib/raw-event-helper-test.js
+++ b/test/javascripts/lib/raw-event-helper-test.js
@@ -12,7 +12,7 @@ module("Unit | Lib | raw-event-helper", function () {
     assert.strictEqual(
       replaceRaw(params, raw),
       'Some text [event param1="newValue1" param2="value2"] more text',
-      "it updates existing parameters and adds new ones"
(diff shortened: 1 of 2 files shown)


test/javascripts/lib/raw-event-helper-test.js
```diff
@@ -0,0 +1,35 @@
...
+    const raw = 'Some text [event param1="value1"] more text';
+    const params = {
+      param1: "newValue1",
+      param2: "value2",
+    };
+
+    assert.strictEqual(
+      replaceRaw(params, raw),
+      'Some text [event param1="newValue1" param2="value2"] more text',
+      "it updates existing parameters and adds new ones"
```
CvX, line 15: for consistency I'd remove "it " prefix from assertion descriptions, like in the one below

CvX (approved):

PR #717, opened 2025-04-10
DEV: update regions

jjaffeux:
This is the result of running `rake javascript:update_constants` to update the definitions.


pmusaraj (approved):
Thanks!

PR #720, opened 2025-04-15
FIX: add line break before and after event block on event insertion

MarineBaron:
If user doesn't well select or put his cursor in composer textarea, the composer text may be malformed after event insertion and it will not create an event after save.

This check composer text selected text and add some line breaks if necessary to avoid malformed text.

**Before fix:**

image
will not create event:
image

**After fix:**
image
will create event
image


davidtaylorhq:
Moving to core repo

PR #725, opened 2025-05-02
DEV: update Indonesia holidays 2025

yuriyaran:
### Done
1. Updated Indonesia public holidays for 2025.

### Would like to get done
1. Tried to DRY `cn_new_lunar_day` (defined in https://github.com/discourse/discourse-calendar/blob/main/vendor/holidays/definitions/hk.yaml#L72) but encountered `make generate` syntax error:
    ```bash
    SyntaxError: /Users/yaran/Projects/discourse-calendar/vendor/holidays/lib/generated_definitions/id.rb:18: syntax error, unexpected ','
    ...ear)", :function_arguments => , :name => "Chinese New Year",...
    ```
2. Automate `INDONESIAN_LUNAR_YEAR_INFO`, so we don't need to update file manually each year


vendor/holidays/reports/coverage/assets/0.12.3/DataTables-1.10.20/images/sort_asc.png
lis2, line 1: Were all those assets added by mistake or they are needed for something?

yuriyaran: well, all those assets were added after running `make update-defs`, as I was following [Release flow](https://github.com/discourse/discourse-calendar/blob/main/vendor/holidays/doc/MAINTAINERS.md#release-flow) directions, which appears to be unnecessary

after discussion with @jjaffeux, I am reverting this change

jjaffeux (approved):

PR #730, opened 2025-05-26
DEV: support for recurrence_until

jjaffeux:
This parameter allows to define the end of the recurrence of an event. For example you can now define an event as every week on Monday, until the 12th of December 2028.

Screenshot 2025-06-02 at 13 39 42


assets/javascripts/discourse/pre-initializers/rich-editor-extension.js
```diff
@@ -12,6 +12,7 @@ const EVENT_ATTRIBUTES = {
   status: { default: "public" },
   timezone: { default: "UTC" },
   allowedGroups: { default: null },
+  recurrenceUntil: { default: null },
```
renato, line 15: Not related to this PR, but I noticed we're missing `recurrence`

PR #761, opened 2025-06-27
FIX: Destroy calendar popover on navigation

martin-brennan:
This fixes an issue where we would show the popover for
a calendar event, but then not destroy it when navigating
to another page, which meant it ended up stuck in the top
left of the page.

This brings the behavior in line with other places we do
`destroyPopover()`, like when the mouse leaves the event
element.

Hovering the event:

image

Where it ended up:

image


martin-brennan:
It would be quite hard to test this since we have no tests already set up for the full calendar on a topic, I think this is pretty low risk and matches existing behaviour

yuriyaran (approved):

PR #764, opened 2025-07-07
DEV: Make specs compatible with upcoming Rails 8 upgrade

Flink:
Rails 8 changes a bit how are handled time objects. When rendering them to a string, they now always include their timezone even for UTC.

### DIFF
--- a/spec/models/discourse_post_event/event_spec.rb
+++ b/spec/models/discourse_post_event/event_spec.rb
@@ -14,10 +14,10 @@ describe DiscoursePostEvent::Event do
     let(:topic) { Fabricate(:topic, user: user) }
     let!(:first_post) { Fabricate(:post, topic: topic) }
     let(:second_post) { Fabricate(:post, topic: topic) }
-    let!(:starts_at) { "2020-04-24 14:15:00" }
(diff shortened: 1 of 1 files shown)


spec/models/discourse_post_event/event_spec.rb
```diff
@@ -49,16 +57,10 @@
...
-              first_post.topic.custom_fields[DiscoursePostEvent::TOPIC_POST_EVENT_STARTS_AT],
-            ).to eq(starts_at)
-            expect(
-              first_post.topic.custom_fields[DiscoursePostEvent::TOPIC_POST_EVENT_ENDS_AT],
-            ).to eq(ends_at)
-            expect(DiscoursePostEvent::EventDate.last.starts_at).to eq_time(
-              DateTime.parse(starts_at),
-            )
-            expect(DiscoursePostEvent::EventDate.last.ends_at).to eq_time(DateTime.parse(ends_at))
+            expect(first_post_starts_at).to eq_time(starts_at)
```
Flink, line 60: That’s the most reliable way to compare time (using time objects), and it works with both Rails 7.2 and 8.0.

cvx (approved):

PR #765, opened 2025-07-08
Update Zimbabwe holidays

Lillinator:
fix Heroe's Day and Defence Forces day

re: https://dev.discourse.org/t/zimbabwe-calendar-definitions/157783

### DIFF
--- a/vendor/holidays/definitions/zw.yaml
+++ b/vendor/holidays/definitions/zw.yaml
@@ -115,14 +115,14 @@ tests:
       
   # Heroes' Day
   - given:
-      date: '2023-08-08'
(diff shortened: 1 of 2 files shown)


cvx (approved):

cvx:
I think you might need to run some tasks? https://github.com/discourse/discourse-calendar?tab=readme-ov-file#holidays

Lillinator:
> I think you might need to run some tasks? https://github.com/discourse/discourse-calendar?tab=readme-ov-file#holidays

oops, yes you are correct!

Lillinator:
does that look correct now?  fixed the tests as well.

cvx (approved):
looks good!

